PR #2, opened 2016-11-04
Use same java executable as parent process by default

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-deployer-local/issues/1

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
@@ -50,12 +50,12 @@ public class LocalDeployerProperties {
 	 * Array of regular expression patterns for environment variables that
 	 * should be passed to launched applications.
 	 */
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -118,4 +118,20 @@ public void setJavaOpts(String javaOpts) {
 		this.javaOpts = javaOpts;
 	}
 
+	public static void main(String[] args) {
```
jvalkeal: I think this main() should go away?

jvalkeal:
Other than that I think it's worth to try this.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
ericbottard: Yes, sorry, this was added as a quick test and I forgot to remove it.

PR #4, opened 2016-11-15
Add profile to deployer configuration

ilayaperumalg:
- This will let us enable/disable specific deployer based on profile


jvalkeal:
Using a profile with autoconfiguration, IMHO, creates more problems than what it's trying to resolve. I'd go with `enable` property. Other than that I don't think a framework lib should use any profiles anyway.

ilayaperumalg:
@jvalkeal ok, by `enable` property do you mean something like `spring.cloud.deployer.local.enable`?

jvalkeal:
yeah, something like that. Think it's a bigger topic to discuss generally how to enable/disable deployers if more that one are in a CP.

ilayaperumalg:
Closing this based on the discussion above. This requires some re-work.

PR #7, opened 2016-11-17
Support deploying docker apps

### LINKED ISSUE #6: Support deploying docker apps
This is the issue to be fixed for https://github.com/spring-cloud/spring-cloud-dataflow/issues/695

ilayaperumalg:
- If the deployment resource is of `DockerResource` then use docker options instead of java options
     - Update the args as environment variables for docker apps

This resolves #6

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -18,7 +18,9 @@ package org.springframework.cloud.deployer.spi.local;
 
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -27,7 +29,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.cglib.core.Local;
+import org.springframework.cloud.deployer.resource.docker.DockerResource;
 import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.core.io.FileSystemResource;
 import org.springframework.http.ResponseEntity;
 import org.springframework.util.Assert;
 import org.springframework.web.client.ResourceAccessException;
@@ -39,6 +43,7 @@ import org.springframework.web.client.RestTemplate;
  *
  * @author Janne Valkealahti
  * @author Mark Fisher
+ * @author Ilayaperumal Gopinathan
  */
 public abstract class AbstractLocalDeployerSupport {
 
@@ -48,6 +53,10 @@ public abstract class AbstractLocalDeployerSupport {
 
 	private final RestTemplate restTemplate = new RestTemplate();
 
+	private final JavaCommandBuilder javaCommandBuilder;
+
+	private final DockerCommandBuilder dockerCommandBuilder;
+
 	/**
 	 * Instantiates a new abstract deployer support.
 	 *
@@ -56,6 +65,8 @@ public abstract class AbstractLocalDeployerSupport {
 	public AbstractLocalDeployerSupport(LocalDeployerProperties properties) {
 		Assert.notNull(properties, "LocalDeployerProperties must not be null");
 		this.properties = properties;
+		this.javaCommandBuilder = new JavaCommandBuilder(properties);
+		this.dockerCommandBuilder = new DockerCommandBuilder();
 	}
 
 	/**
@@ -66,27 +77,6 @@ public abstract class AbstractLocalDeployerSupport {
 	final protected LocalDeployerProperties getLocalDeployerProperties() {
 		return properties;
 	}
-
-	/**
-	 * Builds the execution command for an application.
-	 *
-	 * @param request the request for the application to execute
-	 * @return the build command as a string array
-	 */
-	private String[] buildExecutionCommand(AppDeploymentRequest request) {
-		ArrayList<String> commands = new ArrayList<String>();
-		commands.add(properties.getJavaCmd());
-		Map<String, String> deploymentProperties = request.getDeploymentProperties();
-		ExecutionCommandBuilder commandBuilder = new ExecutionCommandBuilder();
-		// Add Java System Properties (ie -Dmy.prop=val) before main class or -jar
-		commandBuilder.addJavaOptions(commands, deploymentProperties, properties);
-		commandBuilder.addJavaExecutionOptions(commands, request);
-		commands.addAll(request.getCommandlineArguments());
-		logger.debug("Java Commands = " + commands);
-		return commands.toArray(new String[0]);
-	}
-
-
 
 	/**
 	 * Retain the environment variable strings in the provided set indicated by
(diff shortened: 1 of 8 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -73,16 +78,22 @@ final protected LocalDeployerProperties getLocalDeployerProperties() {
...
+			commandBuilder.addDockerOptions(commands, request, args);
+			logger.debug("Docker Command = " + commands);
+		}
+		else {
+			commands.add(properties.getJavaCmd());
+			// Add Java System Properties (ie -Dmy.prop=val) before main class or -jar
+			commandBuilder.addJavaOptions(commands, deploymentProperties, properties);
+			commandBuilder.addJavaExecutionOptions(commands, request);
+			commands.addAll(request.getCommandlineArguments());
+			logger.debug("Java Commands = " + commands);
```
ericbottard: no 's' at `Java Command`

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -73,16 +78,22 @@ final protected LocalDeployerProperties getLocalDeployerProperties() {
...
-		commands.add(properties.getJavaCmd());
 		Map<String, String> deploymentProperties = request.getDeploymentProperties();
 		ExecutionCommandBuilder commandBuilder = new ExecutionCommandBuilder();
-		// Add Java System Properties (ie -Dmy.prop=val) before main class or -jar
-		commandBuilder.addJavaOptions(commands, deploymentProperties, properties);
-		commandBuilder.addJavaExecutionOptions(commands, request);
-		commands.addAll(request.getCommandlineArguments());
-		logger.debug("Java Commands = " + commands);
+		if (request.getResource() instanceof DockerResource) {
+			commandBuilder.addDockerOptions(commands, request, args);
```
ericbottard: Ain't `commands` always empty here?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java
```diff
@@ -90,6 +92,23 @@ public void addJavaExecutionOptions(List<String> commands, AppDeploymentRequest
         }
     }
 
+    public void addDockerOptions(List<String> commands, AppDeploymentRequest request, Map<String, String> args) {
```
ericbottard, line 97: I think this could rather be refactored to be a strategy implementation (decided at `AbstractLocalDeployerSupport#buildExecutionCommand`) and being passed the request and args

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
ilayaperumalg: Yes, That's the `List` we want to return from this method. Do you see an issue with that?

ericbottard: Ah, sorry, did not realize this was modified by side effect. Given the method is currently `void`, maybe it could be returned instead? (along with my other comment about making that class a Strategy as well)

ilayaperumalg:
Squashed and rebased.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -183,4 +180,26 @@ protected static boolean isAlive(Process process) {
...
+//	private class CompositeCommandBuilder implements CommandBuilder {
+//
+//		private final JavaCommandBuilder javaCommandBuilder;
+//
+//		private final DockerCommandBuilder dockerCommandBuilder;
+//
+//		public CompositeCommandBuilder(JavaCommandBuilder javaCommandBuilder, DockerCommandBuilder dockerCommandBuilder) {
+//			this.javaCommandBuilder = javaCommandBuilder;
+//			this.dockerCommandBuilder = dockerCommandBuilder;
+//		}
```
markpollack, line 193: Should remove commented out code

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilderTests.java
```diff
@@ -41,14 +41,14 @@
 
 public class ExecutionCommandBuilderTests {
```
ericbottard, line 42: This is really actually testing `JavaCommandBuilder`, shall rename

ericbottard:
This works well overall except ... we can't undeploy apps :(

This is weird, as a manual `kill` on the spawned docker process indeed triggers shutdown.

ericbottard:
Shouldn't we automatically NAT the `server.port` if present?

ilayaperumalg:
Weird, `undeploy` works for me with the process that runs the `docker run` gets shutdown. Is there anything else left unstopped?

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilderTests.java
ilayaperumalg: Thanks Eric!

ericbottard:
> Shouldn't we automatically NAT the server.port if present?

Actually, I think the `server.port` inside the container should be set to 8080 (or not set) and the computed `server.port` should be NATed to denote a port on the host (similar to what we do for other runtimes).

This may imply that docker support for local should be considered at a level higher than just crafting a commandline and firing it up...

ilayaperumalg:
> Actually, I think the server.port inside the container should be set to 8080 (or not set) and the computed server.port should be NATed to denote a port on the host (similar to what we do for other runtimes).

I am not sure why that matters given everything would be inside the docker container and also, user can provide `server.port` as app property to specify an explicit port.

Now that I think, `undeploy` could possibly invoke `docker stop` though abrupt shutdown of killing the `docker` process (like we do for java exec apps) still works for me and may be ok for `local` deployer?

ericbottard:
> I am not sure why that matters given everything would be inside the docker container

That's why I'm saying it should not be passed to the app, but to the docker process itself, ie

```
docker run -e .... -p 12345:8080 myapp:latest
```

instead of

```
docker run -e server.port=12345 myapp:latest
```

ericbottard:
This would allow graceful shutdown of the app.
This is the equivalent of what we do on CF, mesos, etc

ilayaperumalg:
Closing this PR for now and will re-issue another after some re-work.

ilayaperumalg:
Re-opening this PR after making the following changes and verification:

1) Use `server.port` option as `-p` command in docker run
2) Verified stream undeploy working fine
3) Pass services configuration as deployment properties and verify the apps work together.

ilayaperumalg:
This isn't getting re-opened (may be some permission issue) and the changes pushed into the same branch don't get reflected here as well. Opening a separate PR for the time being.

ilayaperumalg:
May be it's because of the `master` version change after the release.

PR #11, opened 2016-11-25
Support docker apps deployment

### LINKED ISSUE #6: Support deploying docker apps

ilayaperumalg:
- If the deployment resource is of `DockerResource` then use docker options instead of java options
     - Update the args as environment variables for docker apps

This resolves #6

Address review comments

Different approach to set build execution command

Set server.port using -p option


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/JavaCommandBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+		commands.add(properties.getJavaCmd());
+		// Add Java System Properties (ie -Dmy.prop=val) before main class or -jar
+		addJavaOptions(commands, deploymentProperties, properties);
+		addJavaExecutionOptions(commands, request);
+		commands.addAll(request.getCommandlineArguments());
+		logger.debug("Java Command = " + commands);
+		return commands.toArray(new String[0]);
+	}
+
+	protected void addJavaOptions(List<String> commands, Map<String, String> deploymentProperties,
```
ericbottard, line 68: Could this method and the one below be private?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/ExecutionCommandBuilder.java
```diff
@@ -34,6 +35,7 @@
  * A class to help create the execution command for launching a Java Process.
  *
  * @author Mark Pollack
+ * @author Ilayaperumal Gopinathan
  */
 public class ExecutionCommandBuilder {
```
ericbottard: Shouldn't this class be deleted?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/JavaCommandBuilder.java
ilayaperumalg: yeah, I just thought they could be extensible this way. What do you think?

ericbottard:
Squashed and merged as 1acd3a5

PR #12, opened 2016-11-25
Assign qualified bean names for app-deployer/task-laucher

ilayaperumalg:


ilayaperumalg:
This depends on https://github.com/spring-cloud/spring-cloud-deployer/pull/165

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerAutoConfiguration.java
```diff
@@ -35,13 +35,15 @@
 @AutoConfigureOrder(Ordered.HIGHEST_PRECEDENCE)
 public class LocalDeployerAutoConfiguration {
 
-	@Bean
+	public static final String DEPLOYER = "Local";
```
markfisher: Bean names typically begin with a lowercase letter. Is there a reason you want this to start with uppercase?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerAutoConfiguration.java
```diff
@@ -35,13 +35,15 @@
...
+
+	@Bean(DEPLOYER + "AppDeployer")
 	@ConditionalOnMissingBean(AppDeployer.class)
 	public AppDeployer appDeployer(LocalDeployerProperties properties) {
```
markfisher, line 40: maybe we should just handle this by naming the methods `localAppDeployer` and `localTaskLauncher`

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerAutoConfiguration.java
ilayaperumalg: There is no reason for that. will change it to start with `lowerCase`.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerAutoConfiguration.java
ilayaperumalg: I thought about it. Then I thought it would be good to have the `prefix` as a property so that it can be used outside (for instance when choosing the specific deployer based on the derived name)

markfisher: Should we consider using the method name for the bean name but adding a `@Qualifier` value of simply "local"? I guess it depends on the access pattern for these beans.

ilayaperumalg:
@markfisher The usage of `Qualifier` doesn't help specifically when we get the bean from `context`. Also, in order to have multiple deployers in the context, we need to remove the `ConditionalOnMissingBean(AppDeployer.class)` from the app deployer bean methods. Same is the case for task launcher bean methods. I have setup something that incorporates `local` and `cf` deployers in `spring-cloud-cli` launcher here: https://github.com/ilayaperumalg/spring-cloud-cli/commit/39656f98a5494cfda6a1a4fec8855bae6897dcfa

The other approach I am thinking is to exclude `auto-configuration` of the deployers so that we can have only one deployer/task launcher at run time while there can be many in the `classpath`.

The CF deployer tries to connect to CF instance during the deployer auto configuration itself. This means, the excluding the auto configuration of the deployers would work better and there can be only one instance of AppDeployer/TaskLauncher at runtime in the context.

gregturn:
Everyplace I'm using Spring Cloud Deployer CF, I disable autoconfig, because user credentials are requested through the app, hence a connection cannot be created upon startup. Makes me wonder if we can somehow disable connecting with a flag/annotation but keep the rest. ??

ilayaperumalg:
After some discussion, it is decided that we no longer need this approach of keeping unique name per deployer.

BartoszKaszewczuk:
@ilayaperumalg Can you elaborate why so? What was the train of thought behind this conclusion?

I see in [spring-cloud-cli @ 39656f98a5494cfda6a1a4fec8855bae6897dcfa](https://github.com/ilayaperumalg/spring-cloud-cli/commit/39656f98a5494cfda6a1a4fec8855bae6897dcfa) that the solution was to manually fetch `Beans` from context but that doesn't really look like a Spring way of doing things as it still doesn't provide developers with ability to `@Autowire` appropriate implementation depending on the need when having more than one in the context eg `spring-cloud-deployer-local` and `spring-cloud-deployer-cloudfoundry`. Was there some constraint against this approach?
```Java
DeployerProperties deployerProperties = context.getBean(DeployerProperties.class);
final Deployer deployer = context.getBean(Deployer.class);
if (deployerProperties.getTarget().equals(DeployerTarget.local)) {
	deployer.setDeployer(context.getBean(LocalAppDeployer.class));
 }
else if (deployerProperties.getTarget().equals(DeployerTarget.cf)) {
	deployer.setDeployer(context.getBean(CloudFoundryAppDeployer.class));
}
deployer.deploy();
```

ilayaperumalg:
@BartoszKaszewczuk We originally were planning to use Spring Cloud CLI to have multi deployer support. The above conclusion was based on what [Spring Cloud Skipper](https://github.com/spring-cloud/spring-cloud-skipper) provides to support deploying multiple accounts for a given deployer platform and hence the plan to fix this on Spring Cloud CLI is still TBD (Sorry, I should have clarified it better). 

> I see in spring-cloud-cli @ 39656f98a5494cfda6a1a4fec8855bae6897dcfa that the solution was to manually fetch Beans from context but that doesn't really look like a Spring way of doing things as it still doesn't provide developers with ability to @Autowire appropriate implementation depending on the need when having more than one in the context eg spring-cloud-deployer-local and spring-cloud-deployer-cloudfoundry.

I agree. If you think you would like to have the multiple deployer support via Spring Cloud CLI (not via Spring Cloud Skipper) for your use case, please re-open this and we'll plan to address.

BartoszKaszewczuk:
I see. 
The use case behind having each of deployer-implementations annotated with a unique `@Bean` name is for scenarios where developers would need to implement their own solution wrapping a variety of deployers instead of using Spring's solutions.
Eg: Being able to deploy and run the same build of, lets call it a `custom-deployer`, on `CF`, `Kubernetes` and `Local` and have the context deduce which one to use based on some external factors like for example implementations sitting behind a common interface that would Autowire appropriate class containing deployer and some platform specific code by simply using a `@Qualifier(value = ["cf"])` on the dependency of type`TaskLauncher` interface.

Unfortunately I don't think I have permissions to re-open this ticket.

mminella:
Why would you need to use qualified bean names for each deployer type for this approach?  I'd consider it more Spring like to encapsulate that logic in a `FactoryBean` instance that returns the `TaskLauncher` appropriate for the environment.

BartoszKaszewczuk:
`TaskLauncher` beans are already auto-configured by the existing implementation. Encapsulating all the logic within `FactoryBean` would inevitably require much more development effort from the perspective of an external developer, and then that would have to be repeated by other devs with similar use cases. Do you not think it would be beneficial to adjust current implementation of deployers to reduce development time for entities utilising the framework? I'm open to suggestions.

mminella:
Per your use case there still needs to be the logic to "the context deduce which one to use based on some external factors ".  That's what a `FactoryBean` does.  You're asking for "implementations sitting behind a common interface that would Autowire appropriate class containing deployer and some platform specific code".  That's what a `FactoryBean` does.  Are you saying that you'd want the same deliverable to use one deployer in one spot and another in another spot?  If so, that's different.  Chatting about this via Gitter may make more sense than comments on a closed pull request at this point.  Feel free to ping me in the https://gitter.im/spring-cloud/spring-cloud-dataflow channel to continue.

BartoszKaszewczuk:
`FactoryBean` makes sense if the logic for selecting appropriate bean would be more complex, I agree; just like in the snippet from [spring-cloud-cli](https://github.com/ilayaperumalg/spring-cloud-cli/commit/39656f98a5494cfda6a1a4fec8855bae6897dcfa) I posted above. However in simpler scenarios this could also be accomplished by annotating implementations with `@Profile("cf")` or `@Profile("local")` and then simply switching profiles on different envs allowing for autoconf to still happen hence saving dev effort. [Something along these lines (see example)](https://gist.github.com/BartoszKaszewczuk/bbbc7f029c8da6986b7f8d454294aacf). Thoughts?

PR #18, opened 2017-02-20
Add maxConcurrentApp field

shahbour:
Add maxConcurrentApp field that will through an exception if running apps exceed the set number , zero mean unlimited .

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
@@ -42,6 +42,11 @@ public class LocalDeployerProperties {
 	private Path workingDirectoriesRoot = new File(System.getProperty("java.io.tmpdir")).toPath();
 
 	/**
+	 * Maximum number of concurrent Apps allowed to run , zero for infinity
+	 */
+	private int maxConcurrentApps = 0;
+
+	/**
 	 * Whether to delete created files and directories on JVM exit.
 	 */
 	private boolean deleteFilesOnExit = true;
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
Hi @shahbour Thank you for the contribution. Could you share your perspective on how this feature could help especially the `local` deployer is strictly `limited` for `development` environment and not for `production`?

shahbour:
Hello @ilayaperumalg  , i am planning to use local deployer to run some tasks on our production system. Reason for that is we do have our own hardware (storage and vmware) on premises and currently i am not able to install pcf (this is my long term plan for later) so i want to have some kind of queue to control how many task is run at the same time.

sabbyanandan:
Hi, @shahbour. The local-deployer is strictly only meant to be used for development. There's no reliability or resilient capabilities attached to local-deployer. When things fail in production, you will have no opportunity to recover from failures automatically. 

Since this feature aims to make the local-deployer more suitable for production use and given it is outside of this deployer's goals, we will not be able to accept this PR, unfortunately. Please consider migrating to one of the production supported runtime implementations of SCDF. You have CF, K8S, Yarn, or Mesos as options. There are few other community implementations of SCDF, too.

shahbour:
@sabbyanandan thanks the reason i was reluctant in going to SCDF is because my tasks are minimum only couple of tasks a day . i know it is not related but do u have any comparison of the above as i need to go to the lightest one for example pcf recommend 96 GB and this is too much resources for couple of tasks

sabbyanandan:
> i know it is not related but do u have any comparison of the above as i need to go to the lightest one for example pcf recommend 96 GB and this is too much resources for couple of tasks

Assuming you have simple batch-jobs (as tasks) and to run them few times a day, you'd have a trivial resource footprint. The choice of runtime depends on your requirements and organization policies, but I'd think the resource footprint could be more or less similar across all of them. 

Could you clarify what you mean by 96GB?

shahbour:
As i recall i was trying to install pcf on our vmware vshpere but found that the recommended Ram required in 96 GB [requirements](https://docs.pivotal.io/pivotalcf/1-7/customizing/requirements.html)

sabbyanandan:
Thanks, @shahbour. 95G is PCF requirement. If you don't have enough justification to run the Tasks on PCF, perhaps you could consider running it on PWS, too. You'd only be running the apps you need there anyway.  Otherwise, you could provision a small YARN cluster and by default each Task map to 1 CPU core and 1G of RAM - you could override this based on your requirements. Likewise, you could explore other SCDF runtime options as well. 

I'll close this PR. If you have other questions about each runtime, please feel free to reach us at [StackOverflow](http://stackoverflow.com/questions/tagged/spring-cloud-dataflow) or the respective GH [repos](https://github.com/spring-cloud?utf8=%E2%9C%93&q=dataflow-server&type=&language=).

PR #21, opened 2017-03-03
Adapt to changes in TCK

ericbottard:
Fixes https://github.com/spring-cloud/spring-cloud-deployer-local/issues/20

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -153,6 +153,8 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 				}
 			}
 			running.remove(id);
+		} else {
+			throw new IllegalStateException("There is nothing to un-deploy for deployment id " + id);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -153,6 +153,8 @@ public void undeploy(String id) {
 				}
 			}
 			running.remove(id);
+		} else {
+			throw new IllegalStateException("There is nothing to un-deploy for deployment id " + id);
```
ilayaperumalg, line 157: This change needs to be removed as we now have aligned the error message for throwing `IllegalStateException` during `undeploy` across deployer implementations.

ericbottard: I don't understand, I see no special assertion on the message of the exception

ilayaperumalg: There is no special assertion on the message. But, the exception message is changed from this commit: https://github.com/spring-cloud/spring-cloud-deployer-local/commit/34a5bb96d818b4277e677edb800b680fa20846c8 and similar update happened across deployer implementations.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Rebased and merged as 45d9371 after fix the exception message.

PR #26, opened 2017-03-07
Add a note about local-deployer's usage scope

sabbyanandan:


ericbottard:
After second thought, I think this comment belongs in the SCDF core reference manual.
Users are unlikely to get discouraged from using local in production because they read about it here (a building block)

sabbyanandan:
We could add it in both the places.

ericbottard:
Merging this one with formatting changes, please open another PR as you see fit

ericbottard:
Merged as 90f5409b3edba750405a29a55194748b577126e7

PR #38, opened 2017-04-07
Add Metrics system

### LINKED ISSUE #37: Add metrics support

jvalkeal:
- pass in spring.cloud.stream.metrics.key as
  deploymentId + spring.application.index.
- add port and guid to instance attributes
- to easy finding metric instance, if on linux
  also add pid to instance attributes
- Fixes #37

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.deployer.spi.local;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.Field;
 import java.net.HttpURLConnection;
 import java.net.Inet4Address;
 import java.net.URL;
@@ -100,6 +101,7 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 		boolean useDynamicPort = !request.getDefinition().getProperties().containsKey(SERVER_PORT_KEY);
 		HashMap<String, String> args = new HashMap<>();
 		args.putAll(request.getDefinition().getProperties());
+
 		args.put(JMX_DEFAULT_DOMAIN_KEY, deploymentId);
 		if (!request.getDefinition().getProperties().containsKey(ENDPOINTS_SHUTDOWN_ENABLED_KEY)) {
 			args.put(ENDPOINTS_SHUTDOWN_ENABLED_KEY, "true");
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -202,20 +205,27 @@ public void shutdown() throws Exception {
...
+			this.port = port;
 			builder.directory(workDir.toFile());
 			String workDirPath = workDir.toFile().getAbsolutePath();
 			this.stdout = Files.createFile(Paths.get(workDirPath, "stdout_" + instanceNumber + ".log")).toFile();
 			this.stderr = Files.createFile(Paths.get(workDirPath, "stderr_" + instanceNumber + ".log")).toFile();
 			builder.redirectOutput(this.stdout);
 			builder.redirectError(this.stderr);
 			builder.environment().put("INSTANCE_INDEX", Integer.toString(instanceNumber));
-			builder.environment().put("spring.application.index", Integer.toString(instanceNumber));
+			builder.environment().put("spring.application.index", Integer.toString(port));
```
markpollack, line 223: I don't understand why there was a change setting `spring.application.index` from `instanceNumber` to `port`.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -128,6 +130,7 @@ public String deploy(AppDeploymentRequest request) {
 				if (useDynamicPort) {
 					args.put(SERVER_PORT_KEY, String.valueOf(port));
 				}
+				args.put("spring.cloud.stream.metrics.key", deploymentId + ".${spring.application.index}");
```
markpollack, line 133: Maybe this is why the value of `spring.application.index` was changed to port.  Instead, just use the port value that was just calculated directly.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
jvalkeal: This may be oversight from my part but I change it because I think at some point we talked about this index needing to match guid? If so I can change it back as I don't think there is any meaning in metrics. Maybe it was these previous confusions with discussions for index vs. instance.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
jvalkeal: do you want to hardcode port here or use `spring.cloud.application.guid` what I think this ref was supposed to be?

markpollack:
https://github.com/spring-cloud/spring-cloud-deployer-local/pull/39 builds upon this and is the PR for review to be merged.

PR #41, opened 2017-04-12
Add spring profile with repo location to root pom

### LINKED ISSUE #40: Can't cleanly import into IntelliJ

markpollack:
Fixes #40 

The repo locations need to be in the root pom under the spring profile
in order to import the project cleanly into intellij.  Repo location
can't be determined by referencing the parent pom.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,4 +56,56 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+
+	<profiles>
+		<profile>
+			<id>spring</id>
+			<activation><activeByDefault>true</activeByDefault></activation>
+			<repositories>
+				<repository>
+					<id>spring-snapshots</id>
+					<name>Spring Snapshots</name>
+					<url>http://repo.spring.io/libs-snapshot-local</url>
+					<snapshots>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -56,4 +56,56 @@
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
+
+	<profiles>
+		<profile>
+			<id>spring</id>
+			<activation><activeByDefault>true</activeByDefault></activation>
```
ilayaperumalg, line 63: Looking at the other repos [SCDF](https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/pom.xml#L196) and [SCSt](https://github.com/spring-cloud/spring-cloud-stream/blob/master/pom.xml#L190), I don't see the `spring` profile is activated by default. Do you really needed it to be active in order to successfully import in IntelliJ?

trisberg: don't think we need it, I can remove it on merge

trisberg:
Merged as commit 8e87fae590a48a9bdffc9c9b12eb3ba1663872a5

PR #46, opened 2017-05-04
Docker command builder should support custom container names

hekonsek:
Hi,

What do you think about providing some control over the name assigned to created Docker containers? In particular I'd highly appreciate an ability to provide a property which will indicate the name of the created container.

For example for property `docker.container.name=myapp` and `spring.cloud.deployer.count=3`, I'd like to have three containers created with the following names:
```
myapp-0
myapp-1
myapp-2
```

Such addition will be extremely useful for development and test environments, as many existing dev/QA tools and scripts make some assumptions about the containers names.

Cheers!

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.deployer.spi.local;
 import java.net.URL;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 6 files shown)


ericbottard:
Overall, the idea looks good to me, with the exception that `DOCKER_CONTAINER_NAME_KEY` should typically be a **deployment properpty** (ie coming from the `spring.cloud.deployer.xxx` namespace) rather than from `args`. Indeed, it is something that instructs the (specific) deployer how to behave

hekonsek:
Good point, Eric. Added the proper prefix.

ericbottard:
Actually, it's not just a name change. You'll need to change *where* you get the property from.

The aim is that users would ultimately use it like so:
```
stream deploy bo --properties "deployer.log.docker.container.name=gogo"
```

hekonsek:
Ah, sorry - I confused deployment properties with application args. Mea culpa. :) Submitted fix.

hekonsek:
Right now I can specify container name like:

```
Map<String,String> deployProps = new HashMap<>();
deployProps.put("spring.cloud.deployer.docker.container.name", "myName");
Map<String,String> props = new HashMap<>();
deployer.deploy(new AppDeploymentRequest(new AppDefinition("myApp", props), new DockerResource("myContainer"), deployProps));
```

Is that the same result you have in mind?

ericbottard:
Yes, although obviously this would not be done programmatically in the context of Data Flow, but through the `--properties "deployer.foo.docker.container.name=myName"` shell argument.

Your last commit seems legit (haven't tested though).

Thanks again for the contribution!

hekonsek:
Many thanks for a review, Eric. Let me know if I can do anything else to help with this PR.

ericbottard:
Merged as e47ebfb with added test and updated `@author` tags

hekonsek:
Many thanks, Eric!

PR #51, opened 2017-06-02
GH-50 added support for console logging

### LINKED ISSUE #50: Add support for redirecting std.out/err to a parent process's logs
Given that this is a local deployer it does help to have an option to redirect outputs to the parent process's logs (i.e., console) especially in early development stages of where one is testing the deployed app.

olegz:
- added 'inheritLogging' property to LocalDeployerProperties
- added inherited log redirect to LocalAppDeployer if 'inheritLogging' property is set

Resolves #50

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.deployer.spi.local;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.ProcessBuilder.Redirect;
 import java.lang.reflect.Field;
 import java.net.HttpURLConnection;
 import java.net.Inet4Address;
@@ -57,6 +58,7 @@ import org.springframework.util.StringUtils;
  * @author Janne Valkealahti
  * @author Patrick Peralta
  * @author Thomas Risberg
+ * @author Oleg Zhurakousky
  */
 public class LocalAppDeployer extends AbstractLocalDeployerSupport implements AppDeployer {
 
@@ -133,13 +135,16 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 					args.put(SERVER_PORT_KEY, String.valueOf(port));
 				}
 				ProcessBuilder builder = buildProcessBuilder(request, args, Optional.of(i));
-				AppInstance instance = new AppInstance(deploymentId, i, builder, workDir, port);
-				processes.add(instance);
-				if (getLocalDeployerProperties().isDeleteFilesOnExit()) {
-					instance.stdout.deleteOnExit();
-					instance.stderr.deleteOnExit();
+				AppInstance instance = new AppInstance(deploymentId, i, builder, port);
+				if (this.shouldInheritLogging(request)){
+					instance.start(builder);
+					logger.info("Deploying app with deploymentId {} instance {}.\n   Logs will be inherited.", deploymentId, i);
 				}
-				logger.info("Deploying app with deploymentId {} instance {}.\n   Logs will be in {}", deploymentId, i, workDir);
+				else {
+					instance.start(builder, workDir, getLocalDeployerProperties().isDeleteFilesOnExit());
+					logger.info("Deploying app with deploymentId {} instance {}.\n   Logs will be in {}", deploymentId, i, workDir);
+				}
+				processes.add(instance);
 			}
 		}
 		catch (IOException e) {
@@ -190,19 +195,26 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 		}
 	}
 
+	/**
+	 * Will check if {@link LocalDeployerProperties#INHERIT_LOGGING} is set by
+	 * checking deployment properties and then application properties
+	 */
+	private boolean shouldInheritLogging(AppDeploymentRequest request){
+		boolean logToConsole = false;
+		if (request.getDeploymentProperties().containsKey(LocalDeployerProperties.INHERIT_LOGGING)){
+			logToConsole = Boolean.parseBoolean(request.getDeploymentProperties().get(LocalDeployerProperties.INHERIT_LOGGING));
+		}
+		else if (request.getDefinition().getProperties().containsKey(LocalDeployerProperties.INHERIT_LOGGING)){
+			logToConsole = Boolean.parseBoolean(request.getDefinition().getProperties().get(LocalDeployerProperties.INHERIT_LOGGING));
+		}
+		return logToConsole;
+	}
+
 	private static class AppInstance implements Instance, AppInstanceStatus {
 
 		private final String deploymentId;
 
 		private final int instanceNumber;
-
-		private final Process process;
-
-		private final File workFile;
-
-		private final File stdout;
-
-		private final File stderr;
 
 		private final URL baseUrl;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -190,45 +195,51 @@ public void shutdown() throws Exception {
...
+	/**
+	 * Will check if {@link LocalDeployerProperties#INHERIT_LOGGING} is set by
+	 * checking deployment properties and then application properties
+	 */
+	private boolean shouldInheritLogging(AppDeploymentRequest request){
+		boolean logToConsole = false;
+		if (request.getDeploymentProperties().containsKey(LocalDeployerProperties.INHERIT_LOGGING)){
+			logToConsole = Boolean.parseBoolean(request.getDeploymentProperties().get(LocalDeployerProperties.INHERIT_LOGGING));
+		}
+		else if (request.getDefinition().getProperties().containsKey(LocalDeployerProperties.INHERIT_LOGGING)){
```
ericbottard: This is a deployer only concern IMHO, and so should not be checked in the definition

olegz: Actually it is not deployer only concern. The idea is to have two types of control
1. Per deployer
2. Per application

So think of SCDF flow that has bunch of applications and you only care about redirecting logging for one?. . . or you may want all. In the first case it is 'definition' property and the later it is 'deployer'. 
Does it clarify or am I missing something?

ericbottard: I think you are ;)
Per application deployment property is set via 
```
stream deploy foo --properties `deployer.<appname>.<propname>=<value>
```

Application vs Deployment property choice is really that: does this property make sense out of SCDF, if the app was used standalone / without a deployer? If yes, then it may be a case of the former.

Please note that I haven't properly reviewed the PR, so maybe I don't get your "per app / per deployer" split, but to me, this should definitely not be an _application property_ (definition).

olegz:
@ericbottard thanks for the feedback. You're correct that this property is not an app property since it has no influence on the app itself, so yes it is strictly a deployer property in that sense. However.....
My confusion has to do with; how can I pass a deployer property specific to a deployment of a single and specific app (while there may be multiple) and not others?

olegz:
.... and you may have answered my question, so let me review it some more

ericbottard:
I think I have indeed, provided you're not asking how to pass a property to a specific app __instance__ (ie one of the 3 instances of `log`, when you deploy with `deployer.log.count=3`, this is not supported).

If you want to address all the apps that make up a stream foo, you can do so with `deployer.*.<property>=<value>` (works with _app_ properties, as well: `app.*.<property>=<value>`).

Hope that makes sense

olegz:
It does. The per-instances logging redirect may be a nice touch, but the idea behind this enhancement is to primarily facilitate easier debugging/tracing of apps output while in the early stages of the development and IMHO, one most likely not be dealing with multiple instances at that time, and even if there are, the output from all would be sufficient for the cause. In any event we can revisit that one if and when it becomes an issue.
Thanks for looking into this

ericbottard:
Would be interesting to have an option to prepend each line with the
stream/app name (although the default boot logging options include the
pid). BUT not all apps are boot apps.

olegz:
@ericbottard you were indeed correct. Just pushed the latest changes to address your comment 		d13b772. LMK

spring-cloud-deployer-sample-stream/src/main/java/sample/TickTock.java
```diff
@@ -71,8 +63,16 @@ private static AppDeploymentRequest createAppDeploymentRequest(String app, Strin
...
 		}
 		AppDefinition definition = new AppDefinition(app, properties);
-		Map<String, String> environmentProperties = Collections.singletonMap(AppDeployer.GROUP_PROPERTY_KEY, stream);
-		AppDeploymentRequest request = new AppDeploymentRequest(definition, resource, environmentProperties);
+		Map<String, String> deploymentProperties = new HashMap<>();
+		deploymentProperties.put(AppDeployer.GROUP_PROPERTY_KEY, stream);
+		/*
+		 * This will allow output to be logged to the output of the process that started
+		 * the application.
+		 * Could be set per the entire deployment (i.e., 'deployer.inheritLogging=true') or per
```
ericbottard: Not sure this is intended here. Do you mean `deployer.*.inheritLogging`? (with an asterisk?)

olegz: That's exactly what I meant ;). Let me know when you done with review and I'll address it with the last push.

ericbottard: Sorry for the delay. I can fix on merge, don't worry

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -287,6 +298,34 @@ public int getInstanceNumber() {
...
+			if (this.stdout == null) {
+				builder.redirectOutput(Redirect.INHERIT);
+				builder.redirectError(Redirect.INHERIT);
+			}
+			this.process = builder.start();
+		    this.pid = getLocalProcessPid(this.process);
+		}
+
+		private void start(ProcessBuilder builder, Path workDir, boolean deleteOnExist) throws IOException {
+			builder.directory(workDir.toFile());
```
ericbottard: Wondering if the inheriting case should not have a similar line still: logs are one thing, but working directory of the process is another

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -36,6 +37,14 @@
 @ConfigurationProperties(prefix = LocalDeployerProperties.PREFIX)
 public class LocalDeployerProperties {
 
+	/**
+	 * Allows logging to be redirected to the output stream of the process that
+	 * triggered child process.
+	 * Could be set per the entire deployment (i.e., <i>deployer.inheritLogging=true</i>) or per
```
ericbottard: This needs to be fixed as well.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
olegz: @ericbottard  I can go either way on this one, but i don't think it would hurt, so, let me polish some more and I'll make another push

olegz:
@ericbottard just pushed some changes. I think your comments are addressed and I polished the code a bit as well

spring-cloud-deployer-sample-stream/src/main/java/sample/TickTock.java
```diff
@@ -63,8 +63,16 @@ private static AppDeploymentRequest createAppDeploymentRequest(String app, Strin
...
 		}
 		AppDefinition definition = new AppDefinition(app, properties);
-		Map<String, String> environmentProperties = Collections.singletonMap(AppDeployer.GROUP_PROPERTY_KEY, stream);
-		AppDeploymentRequest request = new AppDeploymentRequest(definition, resource, environmentProperties);
+		Map<String, String> deploymentProperties = new HashMap<>();
+		deploymentProperties.put(AppDeployer.GROUP_PROPERTY_KEY, stream);
+		/*
+		 * This will allow output to be logged to the output of the process that started
+		 * the application.
+		 * Could be set per the entire deployment (i.e., 'deployer.*.inheritLogging=true') or per
```
ericbottard, line 71: Will actually remove the `deployer.xx` comments, as they only make sense in the context of SCDF (this is programmatic usage of the spring-cloud-deployer APIs)

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -36,6 +37,14 @@
 @ConfigurationProperties(prefix = LocalDeployerProperties.PREFIX)
 public class LocalDeployerProperties {
 
+	/**
+	 * Allows logging to be redirected to the output stream of the process that
+	 * triggered child process.
+	 * Could be set per the entire deployment (i.e., <i>deployer.*.inheritLogging=true</i>) or per
+	 * individual application (i.e., <i>deployer.&lt;app-name&gt;.inheritLogging=true</i>).
+	 */
+	public static String INHERIT_LOGGING = AppDeployer.PREFIX + "inheritLogging";
```
ericbottard, line 46: Adding `final`

ericbottard, line 46: Also, realizing the prefix should not be `AppDeployer.PREFIX` (ie `spring.cloud.deployer`) but `LocalDeployerProperties.PREFIX` (ie `spring.cloud.deployer.local`)

ericbottard:
Merged as 7f0e064 with extra polishing as seen [here](https://github.com/ericbottard/spring-cloud-deployer-local/commit/12ef43336181b683fbd8bcb2ebe588a1da1cf191)

Issue #55, opened 2017-06-28
NPE observed when querying for runtime status


sabbyanandan:
As an SCDF user, when querying for runtime apps status (with `deployer.*.local.inheritLogging=true`), I see the following NPE. 

**stacktrace:**
```
2017-06-28 10:55:02.567 ERROR 12006 --- [nio-9393-exec-1] o.s.c.d.s.c.RestControllerAdvice         : Caught exception while handling a request

java.lang.NullPointerException: null
    at org.springframework.cloud.deployer.spi.local.LocalAppDeployer$AppInstance.getAttributes(LocalAppDeployer.java:290) ~[spring-cloud-deployer-local-1.2.1.RELEASE.jar!/:1.2.1.RELEASE]
```

**scdf shell:**
```
dataflow:>runtime apps
Command failed org.springframework.cloud.dataflow.rest.client.DataFlowClientException: NullPointerException
```

olegz:
PR is there, but just to follow up as to the cause. . .
The _getAttributes()_ method had a few issues
1. It was re-populating the attributes map every time it is called - redundant puts
2. It was taking the assumption that all/most instance variables are not null and with inheritLogging stdout/stderr are null.

I've restructured the code a bit so now attribute map is populated with values as soon as they are created, thus addressing both 1 and 2.

PR #57, opened 2017-06-28
GH-55 fixed NPE on attribute query

olegz:
Resolves spring-cloud/spring-cloud-deployer-local#55.


ericbottard:
Merged as ea32128 by renaming `result` to `attributes`.

PR #64, opened 2017-07-17
Add escaping on windows

### LINKED ISSUE #63: Escaping issues on windows
When running tests on windows:
```
Results :

Failed tests:   testCommandLineArgumentsPassing(org.springframework.cloud.deployer.spi.local.LocalAppDeployerIntegrationTests): (..)
```
test app fails with message
```
Caused by: java.lang.IllegalArgumentException: Expected 'commandLineArgValueThatMayNeedEscaping' value to be equal to '&'"|< é\(', but was '&'|<'
```

jvalkeal:
- When passing double quotes into processbuilder,
  those needs to be escaped with backslash.
- Moved win detection function to its own utility
  class with protected access as its solely for
  our own usage.
- Fixes #63


trisberg:
LGTM, tests pass on Windows 8.1

PR #68, opened 2017-07-17
Fix local deployer issues

### LINKED ISSUE #62: Support deploying Boot 2.0 apps
Looks like the way we provide the app args by adding them to the environment instead of as command line arguments has some issues with Boot 2.0 based apps. The integration tests set a random port  via `server.port` but the app always ends up using the default `8080`.

### LINKED ISSUE #65: Add test coverage for docker images

### LINKED ISSUE #66: Local deployer doesn't set instance env vars for Docker
The INSTANCE_INDEX env var isn't set when deploying app using docker image

### LINKED ISSUE #67: Local deployer doesn't pass command line args to app when deployed as docker image

trisberg:
- add switch to allow testing with docker images instead of Maven artifacts

- use SPRING_CLOUD_DEPLOYER_SPI_TEST_USE_DOCKER=true

- set instance env vars like INSTANCE_INDEX for app when deployed as docker image

- pass command line args to app when deployed as docker image

- support deploying Boot 2.0 apps by changing to pass app properties as command line args

Resolves #62
Resolves #65
Resolves #66
Resolves #67

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 the original author or authors.
+ * Copyright 2016-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,6 +34,7 @@ import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
  * @author Ilayaperumal Gopinathan
  * @author Eric Bottard
  * @author Henryk Konsek
+ * @author Thomas Risberg
  */
 public class DockerCommandBuilder implements CommandBuilder {
 
@@ -45,26 +46,32 @@ public class DockerCommandBuilder implements CommandBuilder {
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
 	@Override
-	public String[] buildExecutionCommand(AppDeploymentRequest request, Map<String, String> args, Optional<Integer> appInstanceNumber) {
-		List<String> commands = addDockerOptions(request, args, appInstanceNumber);
+	public String[] buildExecutionCommand(AppDeploymentRequest request, Map<String, String> appInstanceEnv,
+										  Map<String, String> appProperties, Optional<Integer> appInstanceNumber) {
+		List<String> commands = addDockerOptions(request, appInstanceEnv, appProperties, appInstanceNumber);
+		// Add appProperties
+		for (String prop : appProperties.keySet()) {
+			commands.add(String.format("--%s=%s", prop, appProperties.get(prop)));
+		}
+		commands.addAll(request.getCommandlineArguments());
 		logger.debug("Docker Command = " + commands);
 		return commands.toArray(new String[0]);
 	}
 
-	private List<String> addDockerOptions(AppDeploymentRequest request, Map<String, String> args, Optional<Integer> appInstanceNumber) {
+	private List<String> addDockerOptions(AppDeploymentRequest request, Map<String, String> appInstanceEnv,
+										  Map<String, String> appProperties, Optional<Integer> appInstanceNumber) {
 		List<String> commands = new ArrayList<>();
 		commands.add("docker");
 		commands.add("run");
-		DockerResource dockerResource = (DockerResource) request.getResource();
-		for (Map.Entry<String, String> entry : args.entrySet()) {
-			if (entry.getKey().equals(LocalAppDeployer.SERVER_PORT_KEY)) {
-				commands.add("-p");
-				commands.add(String.format("%s:8080", args.get(entry.getKey())));
-			}
-			else {
-				commands.add("-e");
-				commands.add(String.format("%s=%s", entry.getKey(), entry.getValue()));
-			}
+		// Add env vars
+		for (String env : appInstanceEnv.keySet()) {
+			commands.add("-e");
+			commands.add(String.format("%s=%s", env, appInstanceEnv.get(env)));
+		}
+		if (appProperties.containsKey(LocalAppDeployer.SERVER_PORT_KEY)) {
+			String port = appProperties.get(LocalAppDeployer.SERVER_PORT_KEY);
+			commands.add("-p");
+			commands.add(String.format("%s:%s", port, port));
 		}
 		if(request.getDeploymentProperties().containsKey(DOCKER_CONTAINER_NAME_KEY)) {
 			if(appInstanceNumber.isPresent()) {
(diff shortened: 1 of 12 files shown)


trisberg:
I tried running the test using docker on my Linux laptop and some tests fail, not sure if this ever worked but I'll see if I can amend the PR if I find the issue

trisberg:
Pushed a fix needed for Docker on Linux, not sure why Mac worked before this fix

trisberg:
Rebased and squashed

markpollack:
I believe there is going to be an issue passing application properties as command line args since command line args have limitations in terms of size.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
```diff
@@ -45,26 +46,32 @@
 	private final Logger logger = LoggerFactory.getLogger(getClass());
 
 	@Override
-	public String[] buildExecutionCommand(AppDeploymentRequest request, Map<String, String> args, Optional<Integer> appInstanceNumber) {
-		List<String> commands = addDockerOptions(request, args, appInstanceNumber);
+	public String[] buildExecutionCommand(AppDeploymentRequest request, Map<String, String> appInstanceEnv,
+										  Map<String, String> appProperties, Optional<Integer> appInstanceNumber) {
+		List<String> commands = addDockerOptions(request, appInstanceEnv, appProperties, appInstanceNumber);
+		// Add appProperties
+		for (String prop : appProperties.keySet()) {
```
markpollack, line 53: I think this should be done instead using `SPRING_APPLICATION_JSON` env-var as the length of commands will get very large and also I believe there are some cases where the relaxed binding syntax of commands will not work, I think this happens for some spring cloud stream binder properties...

trisberg: If the commands would get large, so would the `SPRING_APPLICATION_JSON` env var value. I've never seen any properties that couldn't be passed in on the command line

markpollack: Actually it was the other way around, the relaxed binding gets confused for enviornment variables that bind to maps, but using command line args (with dots) there isn't this confusion.  There still is the issue of length, which might not be so bad - see https://stackoverflow.com/questions/6846263/maximum-length-of-command-line-argument-that-can-be-passed-to-sqlplus-from-lin

which for me gives 2MB of command line args on ubuntu, but it is significantly smaller for windows that we should look to using a different approach

https://support.microsoft.com/en-us/help/830473/command-prompt-cmd--exe-command-line-string-limitation

https://stackoverflow.com/questions/3205027/maximum-length-of-command-line-string

either 8KB or 32KB....

trisberg: The CF deployer has a `spring.cloud.deployer.cloudfoundry.use-spring-application-json` property, maybe we can have the same here and default to `true`

trisberg: is there a limit for environment variables?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -67,7 +67,8 @@
 	 * Array of regular expression patterns for environment variables that
 	 * should be passed to launched applications.
 	 */
-	private String[] envVarsToInherit = {"TMP", "LANG", "LANGUAGE", "LC_.*", "PATH"};
+	private String[] envVarsToInherit = {"TMP", "LANG", "LANGUAGE", "LC_.*", "PATH",
+			"SPRING_CLOUD_APPLICATION_GUID", "SPRING_APPLICATION_INDEX", "INSTANCE_INDEX"};
```
jvalkeal: Why this change? These are set by deployer and should not inherit from a parent process even in cases where deployer doesn't set it.

jvalkeal:
Looks like we don't get into trouble with escaping which I think is cause by `ProcessLauncher` working so low level that we don't get into same trouble than with bash. This was a nice surprise.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
jvalkeal: btw, on windows we don't execute via cmd, java's `ProcessBuilder` uses its own win api's which hopefully have different limitation. This is a common complain on win as low level limits are rather high but then every higher level win api may have their own limits.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
trisberg: this was necessary due to a change where we moved adding these env vars earlier in the process instead of setting them directly on the process builder (which didn't work for the docker scenario since the env vars weren't passed in to the container). Also, the way the env vars are handled is subtractive, so unless we add these env vars to `envVarsToInherit` they get removed. Maybe we should rename the variable to be `envVarsToKeep`

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
markpollack: This post doesn't inspire confidence either...

https://blogs.msdn.microsoft.com/oldnewthing/20100203-00/?p=15083

"If you have to ask what the limit is, you're doing it wrong"

markpollack: Also, since the local deployer is going to be used in replacing spring batch admin, there are probably batch jobs (which aren't boot) and expect to see something passed in via the command line.  @mminella can you comment?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
jvalkeal: Right ok gotcha, think I understand it now.

trisberg: Looking at it again, we should move the deployer provided env vars to a different place, they don't really belong in the deployer properties

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupportTests.java
```diff
@@ -0,0 +1,90 @@
...
+
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
+
+/**
+ * Tests for the AbstractLocalDeployerSupport
+ *
+ * @author Thomas Risberg
+ */
+public class AbstractLocalDeployerSupportTests {
```
markpollack, line 38: It isn't an abstract class....

markpollack:
Removed `Abstract` prefix from test, squashed and merged as e7cceab5ad39a8cd625bbd797e017baa20ff3e74

PR #69, opened 2017-07-19
Updated README to allow for production use

mminella:
This commit rewords the README to read that production use is ok within
the given parameters of what the local deployer is capable of doing.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,8 +1,17 @@
 == Spring Cloud Local Deployer
 
-WARNING: Spring Cloud Local Deployer is strictly only meant to be used for development. This 
-implementation is not recommended for production use because there’s no high availability, fault-tolerance, 
-or resiliency mechanics; hence, there are no reliable execution guarantees associated with it.
+Spring Cloud Local Deployer is an implementation of the Spring Cloud Deployer SPI for use
+to deploy applications on the same machine.  This occurs by this application spawning a
+new JVM process for the deployed application.
+
+NOTE:  It's important to note that this deployer spawns new JVMs that are not monitored
+or maintained by this deployer.  No attempts at high availability, fault tolerance, or
+resiliency are provided by the deployer.  Since the deployer SPI expects an underlying
+platform to provide that level of resiliency, any use of this deployer in a production
+environment should be accompanied with additional monitoring at the app level (the apps
+this deployer deploys).  This deployer will not be updated to take on those requirements.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -1,8 +1,14 @@
...
-or resiliency mechanics; hence, there are no reliable execution guarantees associated with it.
+Spring Cloud Local Deployer is an implementation of the Spring Cloud Deployer SPI for use
+to deploy applications on the same machine.  This occurs by this application spawning a
+new JVM process for the deployed application.
+
+NOTE:  It's important to note that this deployer spawns new JVMs that are not monitored
+or maintained by this deployer.  No attempts at high availability, fault tolerance, or
+resiliency are provided by the deployer.  Any use of this deployer in a production
+environment should be accompanied with additional monitoring at the app level (the apps
+this deployer deploys) to provide production grade resiliency.
```
markfisher: might want to emphasize that we do not intend to add such features for this deployer at any point in the future, and that in general, we recommend using environments (cloud foundry, k8s, etc) where such features are enabled out-of-the-box

mminella: How about this (I'll update the PR if it's ok):
```
NOTE:  It's important to note that this deployer spawns new JVMs that are not monitored
or maintained by this deployer.  No attempts at high availability, fault tolerance, or
resiliency are provided by the deployer.  Since the deployer SPI expects an underlying
platform to provide that level of resiliency, any use of this deployer in a production
environment should be accompanied with additional monitoring at the app level (the apps
this deployer deploys).  This deployer will not be updated to take on those requirements
and encourage users to explore the CloudFoundry and Kubernetes variants as ways to meet
them.
```

trisberg: Maybe add a WE in the last sentence like: "This deployer will not be updated to take on those requirements and WE encourage users to explore the CloudFoundry and Kubernetes variants as ways to meet them."

mminella: I only held that out because I wasn't sure who the "we" was in a README...

markfisher: then you could say: "This deployer will not be updated to take on those requirements. Therefore the user is encouraged to explore the CloudFoundry and Kubernetes variants as ways to meet them."

mminella: If everyone is ok with that, I'll update the PR...

trisberg: +1

mminella: Updated

trisberg:
LGTM

PR #71, opened 2017-07-24
Change random dir length on windows tests

### LINKED ISSUE #70: Dir length on win causes trouble in CI
When deployment directory is created on windows with tests, use of multiple UUID's causes long paths which seem to fail `ProcessBuilder`.

jvalkeal:
NOTE: I ran CI build against my own branch with these changes and build was green.

- Created directory path is too long due to
  use of multiple UUID's causing ProcessBuilder
  failure on windows. Change random name to be
  shorter on windows using less reliable
  shorter UUID.

Fixes #70
Fixes spring-cloud/spring-cloud-dataflow#1586


ghillert:
* Added Author tags
* Rebased

Merged with commit b4f34628d57a7a253cf8e56c39edcd8cafee1cdb

PR #80, opened 2017-11-06
Adds check for server.port in the command args

### LINKED ISSUE #77: TaskLauncher adds server.port eventhough it is specified in args
The task deployer checks to see if the server.port is in the properties https://github.com/spring-cloud/spring-cloud-deployer-local/blob/master/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java#L89.  But does not check to see if it has been added to the arguments list.

cppwfs:
resolves #77


markpollack:
reversed some of the if/else logic so it reads a bit more cleanly.

squashed and merged as 2f6495dc4a4696bb46171966f80bb677087fb957

PR #85, opened 2018-01-09
Added remote debugging configuration

### LINKED ISSUE #35: Add support to debug deployed/launched applications

olegz:
Added remote debugging configuration to LocalDeployer which coudl now be set using 'debugPort' and 'debugSuspend' properties

Resolves #35

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -137,6 +137,11 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 				Map<String, String> appInstanceEnv = new HashMap<>();
 				AppInstance instance = new AppInstance(deploymentId, i, appInstanceEnv, port);
 				ProcessBuilder builder = buildProcessBuilder(request, appInstanceEnv, args, Optional.of(i)).inheritIO();
+
+				if (request.getDeploymentProperties().containsKey(LocalDeployerProperties.DEBUG_PORT)) {
+					builder.command().add(1, this.buildRemoteDebugInstruction(request.getDeploymentProperties()));
+				}
+
 				builder.directory(workDir.toFile());
 				if (this.shouldInheritLogging(request)){
 					instance.start(builder, workDir);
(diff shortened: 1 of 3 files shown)


sabbyanandan:
Hi, @aclement. A _very long_ time ago, you had tried to add similar support for it via: spring-cloud/spring-cloud-dataflow#369. When you get a chance, it'd be great to hear your thoughts on this.

olegz:
@aclement to simplify the review here is the doc PR hat explains its usage within data-flow https://github.com/spring-cloud/spring-cloud-dataflow/pull/1909/commits/584d148c220d1addd4d47bf7d845ab7b2d0c72b5

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -137,6 +137,11 @@ public String deploy(AppDeploymentRequest request) {
 				Map<String, String> appInstanceEnv = new HashMap<>();
 				AppInstance instance = new AppInstance(deploymentId, i, appInstanceEnv, port);
 				ProcessBuilder builder = buildProcessBuilder(request, appInstanceEnv, args, Optional.of(i)).inheritIO();
+
+				if (request.getDeploymentProperties().containsKey(LocalDeployerProperties.DEBUG_PORT)) {
+					builder.command().add(1, this.buildRemoteDebugInstruction(request.getDeploymentProperties()));
```
markpollack: In the original PR from andy there was this tidbit 
```
command.add("-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address="+(debugPort+i));
```
The code is in a loop to deploy the apps based on the deployment property `count` so the port number needs to be difference for each instance that is deployed.  That is they there is the `+i` off the "basePort".

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -197,6 +202,23 @@ public void shutdown() throws Exception {
 		}
 	}
 
+	private String buildRemoteDebugInstruction(Map<String, String> deploymentProperties) {
+		String dp = deploymentProperties.get(LocalDeployerProperties.DEBUG_PORT);
+		String ds = deploymentProperties.getOrDefault(LocalDeployerProperties.DEBUG_SUSPEND, "y");
+		StringBuilder debugCommandBuilder = new StringBuilder();
+		if (StringUtils.hasText(dp)) {
+			logger.warn("***** [REMOTE DEBUGGING] REMOTE DEBUGGING IS ENABLED ON PORT " + dp + " *****");
```
markpollack: The `deploymentId` should be passed into this method to give context to these warning messages, otherwise one can't tell what app is having remote debugging enabled. I think we should town down the `ALL CAPS` and `****` as well.

markpollack: Actually, while doing some refactoring, this property should also be available for when deploying Tasks....so it should move up into the `AbstractLocalDeployerSupport` class and `LocalTaskLauncher` can take advantage of it as well.

markpollack: I'll push a variation on this to a new branch.

olegz:
#86 includes it

PR #90, opened 2018-02-05
Fixed SAJ not being used by local and docker deployers

### LINKED ISSUE #88: DeployerLocal adds all environment variables to command Line
When attempting to run a task from the command line, all environment variables are added to the command line. 
This causes 2 problems: 
1) if the entry has no value boot throws an IllegalArgumentException as discussed here https://github.com/spring-cloud/spring-cloud-task/issues/369
2) If the user enumerates the datasource via environment properties it is not added to the command properly, because it uses underscores instead of dots.

While this solution may work for Docker, it does not work for running from the command line.

viniciusccarvalho:
Fixes #88
 - SAJ was not being used at all for the deployers, fixed missing
   environment variable copy
 - Added a couple of tests to test the ProcessBuilder itself

### DIFF
--- a/spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilderTests.java
+++ b/spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilderTests.java
@@ -17,14 +17,14 @@
 package org.springframework.cloud.deployer.spi.local;
 
 import static org.hamcrest.collection.IsArrayContainingInOrder.arrayContaining;
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
 
-import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 
-import org.hamcrest.collection.IsArrayContainingInOrder;
 import org.junit.Test;
 
 import org.springframework.cloud.deployer.resource.docker.DockerResource;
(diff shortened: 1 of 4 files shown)


spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/JavaExecutionCommandBuilderTests.java
```diff
@@ -164,6 +168,26 @@ public void testMissingMain() throws MalformedURLException {
...
+
+        Map<String, String> deploymentProperties = new HashMap<>();
+        deploymentProperties.put(LocalDeployerProperties.DEBUG_PORT, "9999");
+        deploymentProperties.put(LocalDeployerProperties.DEBUG_SUSPEND, "y");
+        deploymentProperties.put(LocalDeployerProperties.INHERIT_LOGGING, "true");
+        AppDeploymentRequest request = new AppDeploymentRequest(definition, testResource(), deploymentProperties);
+
+
+        ProcessBuilder builder = deployer.buildProcessBuilder(request, Collections.emptyMap(), request.getDefinition().getProperties(), Optional.of(1), "foo" );
+        assertThat(builder.environment().keySet(), hasItem("SPRING_APPLICATION_JSON"));
```
markpollack, line 187: should assert that foo=bar is in there.

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilderTests.java
```diff
@@ -17,14 +17,14 @@
 package org.springframework.cloud.deployer.spi.local;
 
 import static org.hamcrest.collection.IsArrayContainingInOrder.arrayContaining;
+import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
```
mminella, line 21: Why aren't these expanded out?  Also, I know Framework's rules are to have static imports at the bottom of the import list: https://github.com/spring-projects/spring-framework/wiki/Code-Style#import-statements

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/JavaExecutionCommandBuilderTests.java
```diff
@@ -27,15 +28,18 @@
 
 import static org.hamcrest.Matchers.containsInAnyOrder;
 import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.is;
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
```
mminella, line 34: Same her on the static imports.  At a minimum, they shouldn't be nested in the middle of the imports.

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilderTests.java
markpollack: we don't have checkstyle in place for this project....so just depends on local IDE settings.

I've merged this already, I can go in and optimize the imports and re-commit

markpollack:
added additional assert and optimized imports across the code base

* bce3992e5fb2126c845db09582d2f58772944655 optimize imports
* c463b4b63fadf366c7b74ef7395869ae6094e5af Fixed SAJ not being used by local and docker deployers

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/JavaExecutionCommandBuilderTests.java
viniciusccarvalho: Probably my IDE is not configured, I always run `mvn install -Pfull` which usually what triggers the checkstyle plugin, I was not aware it was not configured here

PR #92, opened 2018-02-12
Set SPRING_APPLICATION_JSON as default

### LINKED ISSUE #91: Set SPRING_APPLICATION_JSON as the default mode to pass properties for running apps
Switch to `SPRING_APPLICATION_JSON` mode to pass application properties to apps
Make sure that if caller has already passed a `SPRING_APPLICATION_JSON` variable, we merge it with what the deployer passes to the application.

viniciusccarvalho:
Fixes #91
- SPRING_APPLICATION_JSON is now default way to pass properties
- If present as application properties, we merge it

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -226,8 +226,6 @@ public abstract class AbstractLocalDeployerSupport {
 				}
 				appInstanceEnvToUse.putAll(Collections.singletonMap(
 						SPRING_APPLICATION_JSON, OBJECT_MAPPER.writeValueAsString(localApplicationProperties)));
-			} catch (JsonProcessingException e) {
-				throw new RuntimeException(e);
 			}
 			catch (IOException e) {
 				throw new RuntimeException(e);
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.deployer.spi.local;
 
+import java.io.IOException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -29,6 +30,7 @@ import java.util.Set;
 import java.util.regex.Pattern;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -50,11 +52,14 @@ import org.springframework.web.client.RestTemplate;
  * @author Ilayaperumal Gopinathan
  * @author Thomas Risberg
  * @author Oleg Zhurakousky
+ * @author Vinicius Carvalho
  */
 public abstract class AbstractLocalDeployerSupport {
 
 	public static final String USE_SPRING_APPLICATION_JSON_KEY =
 			LocalDeployerProperties.PREFIX + ".use-spring-application-json";
+
+	public static final String SPRING_APPLICATION_JSON = "SPRING_APPLICATION_JSON";
 
 	private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -213,11 +218,20 @@ protected void handleAppPropertiesPassing(AppDeploymentRequest request, Map<Stri
...
+				}
 				appInstanceEnvToUse.putAll(Collections.singletonMap(
-						"SPRING_APPLICATION_JSON", OBJECT_MAPPER.writeValueAsString(appProperties)));
+						SPRING_APPLICATION_JSON, OBJECT_MAPPER.writeValueAsString(localApplicationProperties)));
 			} catch (JsonProcessingException e) {
 				throw new RuntimeException(e);
 			}
+			catch (IOException e) {
+				throw new RuntimeException(e);
+			}
```
mminella: Wrapping these two in just a `RuntimeException` feels a bit lax.  Should we create a custom exception for this?  Also, if this project requires java 8, we can use the multiple exception syntax:
```
catch(JsonProcessingException | IOException ex){
```

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -50,12 +52,15 @@
...
  * @author Thomas Risberg
  * @author Oleg Zhurakousky
+ * @author Vinicius Carvalho
  */
 public abstract class AbstractLocalDeployerSupport {
 
 	public static final String USE_SPRING_APPLICATION_JSON_KEY =
 			LocalDeployerProperties.PREFIX + ".use-spring-application-json";
 
+	public static final String SPRING_APPLICATION_JSON = "SPRING_APPLICATION_JSON";
```
mminella, line 62: I know this isn't a short term option, but since Boot doesn't expose this as a constant that we can consume, we may want to consider moving this to the SPI project so it's consistent across the deployers.

viniciusccarvalho: 100% agreeing with you, but since touching the SPI was something a bit more complicated, I left it behind, just moved to a constant to avoid typos seemed logical

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
viniciusccarvalho: Multiple exception syntax is a no brainer for sure, forgot this was a java8 project. Adding a new Exception, maybe at the `buildProcessBuilder`, would be nice if the `AppDeployer` had some sort of `DeploymentFailedException` so at least we could base out of that. Creating an exception to be caught at the LocalAppDeployer to propagate `RuntimeException` as it happens for other things, seems like that would not be the right place. Another example where we may need to touch the SPI.

@jvalkeal made a comment recently on how the k8s deployer needs some love, I'd say they all need some attention to get to a more cohesive state. They all obey the spec of course, but we could benefit from some patterns that happen on all different deployers (the EntryPoint style is one of many)

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
mminella: Actually...change this comment.  I asked Boot to create a constant for this value and they have.  Once available, that would be the best route IMHO.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
viniciusccarvalho: BTW we can't have multiple catch syntax on this, because JsonProcessingException is a subclass of IOException.

mminella: Then why have both catches in the first place?

viniciusccarvalho: I'm fixing this, I'm not saying we need both either. This was just the IDE suggesting the inclusion of IOException, where we had `JsonProcessingException` already in place.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -106,7 +107,7 @@
 	 * Flag to indicate whether application properties are passed as command line args or in a
 	 * SPRING_APPLICATION_JSON environment variable.
 	 */
-	private boolean useSpringApplicationJson = false;
+	private boolean useSpringApplicationJson = true;
```
markpollack, line 110: Added Javadoc to indicate the default value is true.

markpollack:
I get two failing tests `LocalDeployerSupportTests.testAppPropsAsCommandLineArgs` and `LocalDeployerSupportTests.testSpringApplicationJson`

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/JavaExecutionCommandBuilderTests.java
```diff
@@ -182,6 +181,28 @@ public void testCommandBuilderSpringApplicationJson() throws Exception {
...
+        AppDefinition definition = new AppDefinition("foo", applicationProperties);
+
+        deploymentProperties.put(LocalDeployerProperties.DEBUG_PORT, "9999");
+        deploymentProperties.put(LocalDeployerProperties.DEBUG_SUSPEND, "y");
+        deploymentProperties.put(LocalDeployerProperties.INHERIT_LOGGING, "true");
+        AppDeploymentRequest request = new AppDeploymentRequest(definition, testResource(), deploymentProperties);
+
+
+        ProcessBuilder builder = deployer.buildProcessBuilder(request, Collections.emptyMap(), request.getDefinition().getProperties(), Optional.of(1), "foo" );
+        assertThat(builder.environment().keySet(), hasItem("SPRING_APPLICATION_JSON"));
```
markpollack, line 202: can use the contant in AbstractLocalDeployerSupport

markpollack:
something strange happened, not sure why it got closed.

markpollack:
merged as d24039a3d2d5be3a3ae3c350fd6aa15d75fb63c6

PR #96, opened 2018-02-27
Fixes wrong test assertion with SAJ enabled

### LINKED ISSUE #91: Set SPRING_APPLICATION_JSON as the default mode to pass properties for running apps

viniciusccarvalho:
Supposed to fix a erroneous test that changed behavior on #91. Should address concerns expressed in #92 

Resolves #91


markpollack:
This was already merged as part of d24039a3d2d5be3a3ae3c350fd6aa15d75fb63c6

PR #97, opened 2018-02-28
Better handling of env variables on windows

### LINKED ISSUE #95: Inherit ENV variables in Windows OS
_From @VovaViter on January 31, 2018 14:2_

Hi All.

I spent much time when try finding problems with inheritance ENV variables such as PATH on Windows 10. 
Please fix this problem in LocalDeployerProperties and change envVarsToInherit array according to the OS. On  Windows 10 ENV variables written in camel case. 

_Copied from original issue: spring-cloud/spring-cloud-dataflow#2017_

jvalkeal:
- Use different defaults on windows what env
  variables to inherit on default.
- Add new test via test app to check via
  boot's env endpoint what actually got set.
- Fixes #95


ghillert:
LGTM - merged with commit c5e93c4eb666b5098ea2317b5d99501d7a146157

PR #105, opened 2018-03-14
Refactor server port calculation

mminella:
This commit changes the way the server port calculation is performed.
In order of precidence, it will calculate a dynamic port if required,
otherwise it will obtain the server port from the command line args,
otherwise it will obtain the server port from the application
properties, and if none of those options are valid 8080 will be
returned.

Resolves spring-cloud/spring-cloud-deployer-local#106

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -54,6 +54,10 @@ public abstract class AbstractLocalDeployerSupport {
 	private static final String USE_SPRING_APPLICATION_JSON_KEY =
 			LocalDeployerProperties.PREFIX + ".use-spring-application-json";
 
+	public static final String SERVER_PORT_KEY = "server.port";
+
+	public static final String SERVER_PORT_KEY_PREFIX = "--" + SERVER_PORT_KEY + "=";
+
 	public static final String SPRING_APPLICATION_JSON = "SPRING_APPLICATION_JSON";
 
 	private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
@@ -68,7 +72,7 @@ public abstract class AbstractLocalDeployerSupport {
 
 	private final DockerCommandBuilder dockerCommandBuilder;
 
-	private static final int DEFAULT_SERVER_PORT = 8080;
+	public static final int DEFAULT_SERVER_PORT = 8080;
 
 	private String[] envVarsSetByDeployer =
 			{"SPRING_CLOUD_APPLICATION_GUID", "SPRING_APPLICATION_INDEX", "INSTANCE_INDEX"};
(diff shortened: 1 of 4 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -54,6 +54,10 @@
 	private static final String USE_SPRING_APPLICATION_JSON_KEY =
 			LocalDeployerProperties.PREFIX + ".use-spring-application-json";
 
+	public static final String SERVER_PORT_KEY = "server.port";
+
+	public static final String SERVER_PORT_KEY_PREFIX = "--" + SERVER_PORT_KEY + "=";
```
markpollack, line 59: The "prefix" prepending `--` and adding `=` seems a bit odd.  `SERVER_PORT_KEY_COMMAND_LINE_ARG` instead?

markpollack: We could also keep them at package level vs. public

mminella: I just moved the existing field, but I agree that the naming doesn't make sense...I can fix if you want.

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncherIntegrationTests.java
```diff
@@ -80,6 +94,31 @@ protected String randomName() {
...
+		Map<String, String> appProperties = new HashMap();
+		appProperties.put("killDelay", "0");
+		appProperties.put("exitCode", "0");
+
+		AppDefinition definition = new AppDefinition(this.randomName(), appProperties);
+
+		Resource resource = this.testApplication();
+
+		List<String> commandLineArgs = new ArrayList<>(1);
+		commandLineArgs.add(LocalTaskLauncher.SERVER_PORT_KEY_PREFIX + SocketUtils.findAvailableTcpPort(LocalTaskLauncher.DEFAULT_SERVER_PORT));
```
markpollack, line 108: This test doesn't really validate that the app got the `server.port`....just that it didn't crash.  Commenting the line out makes this test equivalent to `testSimpleLaunch` in AbstractTaskLauncherIntegrationTests.  Prob would need to hit the actuator endpoint to make sure the `server.port` value is what was expectedat an integration test level.    if we keep it, maybe the one line should just be added to `testSimpleLaunch`

mminella: There probably isn't Actuator in the app which is why I didn't go further (if I'm wrong, then I agree, but we usually don't include it in tasks).  The test does fail in the way the build was failing when the old code is in place (the new code fixes the test).

markpollack: I'll add a unit test for calcServerPort to ensure the behavior/overides and keep this one.  Moving it up to AbstractTaskLauncherIntegration tests may end up breaking it on other platforms which we don't need to deal with now.  I'll create an issue to investigate moving it to the base class.

mminella: Is this base class used via the other deployers?  I was running under the assumption that code common across deployer impls would live in the SPI project...

markpollack:
Added additional test, reverted exposing some static fields as public.

Merged as a4d623b7ac781fb903e3a5c5449becdc2ee38129

PR #110, opened 2018-03-28
Fix property handling with SAJ

### LINKED ISSUE #109: Metrics broken with collector
This is a regression originally created into `spring-cloud-stream` and then in spring-cloud/spring-cloud-dataflow#2118. Seem to originate from #91 with d24039a3d2d5be3a3ae3c350fd6aa15d75fb63c6.

Everything were moved into `SPRING_APPLICATION_JSON` which itself is wrong as options passed to deployer to get added into environment should go into environment and not into `SPRING_APPLICATION_JSON`.

With boot2 apps we catched this immediately as apps failed during a startup. Boot1 apps start ok but metrics are broken because `spring.cloud.stream.metrics.key` is wrong.

```
2018-03-27 12:03:03.201 DEBUG 29132 --- [jO5qL_ytbUlVQ-1] o.s.integration.channel.DirectChannel    : postSend (sent=true) on channel 'input', message: GenericMessage [payload={"name":"foostream2.log131.${spring.cloud.application.guid}","metrics":[{"name":"integration.channel.input.errorRate.mean","value":0.0,"timestamp":"2018-03-27T11:03:03.198Z"}
```
We can see from below json that `"SPRING_CLOUD_APPLICATION_GUID\":\"40302\"` is not mapped to `"spring.cloud.stream.metrics.key\":\"foostream2.log131.${spring.cloud.application.guid}\"` because boot would resolve it from environment variables, as deployer should set it, not from json itself.
```
SPRING_APPLICATION_JSON": "{\"endpoints.jmx.unique-names\":\"true\",\"spring.cloud.dataflow.stream.app.label\":\"log131\",\"endpoints.shutdown.enabled\":\"true\",\"spring.cloud.stream.metrics.properties\":\"spring.application.name,spring.application.index,spring.cloud.application.*,spring.cloud.dataflow.*\",\"SPRING_CLOUD_APPLICATION_GUID\":\"40302\",\"spring.cloud.stream.bindings.applicationMetrics.destination\":\"metrics\",\"INSTANCE_INDEX\":\"0\",\"security.basic.enabled\":\"false\",\"SPRING_APPLICATION_INDEX\":\"0\",\"spring.cloud.dataflow.stream.name\":\"foostream2\",\"endpoints.configprops.keys-to-sanitize\":\"blaah\",\"spring.metrics.export.triggers.application.includes\":\"integration**\",\"spring.cloud.stream.metrics.key\":\"foostream2.log131.${spring.cloud.application.guid}\",\"spring.cloud.stream.bindings.input.group\":\"foostream2\",\"management.security.enabled\":\"false\",\"server.port\":\"40302\",\"spring.cloud.dataflow.stream.app.type\":\"sink\",\"spring.cloud.stream.bindings.input.destination\":\"foostream2.time131\",\"endpoints.env.keys-to-sanitize\":\"blaah\",\"spring.jmx.default-domain\":\"foostream2.log131\",\"spring.cloud.application.group\":\"foostream2\"}"
```

Similarly UI doesn't show any metrics because this key is wrong.

jvalkeal:
- Now as a quick workaround to get metrics working,
  conditionally add normal properties reflecting i.e.
  SPRING_CLOUD_APPLICATION_GUID as spring.cloud.application.guid.
- See from added comments in code why it wasn't possible to
  remove those env style formats.
- This change now allows scdf server to show metrics again.
- Fixes #109

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -171,19 +171,15 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 
 				int port = calcServerPort(request, useDynamicPort, appInstanceEnv);
 
-				appInstanceEnv.put("INSTANCE_INDEX", Integer.toString(i));
-				appInstanceEnv.put("SPRING_APPLICATION_INDEX", Integer.toString(i));
-				appInstanceEnv.put("SPRING_CLOUD_APPLICATION_GUID", Integer.toString(port));
-
-				// we only set 'normal' style props reflecting what we set for env format
-				// for cross reference to work inside SAJ.
-				// looks like for now we can't remove these env style formats as i.e.
-				// DeployerIntegrationTestProperties in tests really assume 'INSTANCE_INDEX' and
-				// this might be indication that we can't yet fully remove those.
 				if (useSpringApplicationJson(request)) {
 					appInstanceEnv.put("instance.index", Integer.toString(i));
 					appInstanceEnv.put("spring.application.index", Integer.toString(i));
 					appInstanceEnv.put("spring.cloud.application.guid", Integer.toString(port));
+				}
+				else {
+					appInstanceEnv.put("INSTANCE_INDEX", Integer.toString(i));
+					appInstanceEnv.put("SPRING_APPLICATION_INDEX", Integer.toString(i));
+					appInstanceEnv.put("SPRING_CLOUD_APPLICATION_GUID", Integer.toString(port));
 				}
 
 				AppInstance instance = new AppInstance(deploymentId, i, port);
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
@@ -175,6 +175,17 @@ public class LocalAppDeployer extends AbstractLocalDeployerSupport implements Ap
 				appInstanceEnv.put("SPRING_APPLICATION_INDEX", Integer.toString(i));
 				appInstanceEnv.put("SPRING_CLOUD_APPLICATION_GUID", Integer.toString(port));
 
+				// we only set 'normal' style props reflecting what we set for env format
+				// for cross reference to work inside SAJ.
+				// looks like for now we can't remove these env style formats as i.e.
+				// DeployerIntegrationTestProperties in tests really assume 'INSTANCE_INDEX' and
+				// this might be indication that we can't yet fully remove those.
+				if (useSpringApplicationJson(request)) {
+					appInstanceEnv.put("instance.index", Integer.toString(i));
+					appInstanceEnv.put("spring.application.index", Integer.toString(i));
+					appInstanceEnv.put("spring.cloud.application.guid", Integer.toString(port));
+				}
+
 				AppInstance instance = new AppInstance(deploymentId, i, port);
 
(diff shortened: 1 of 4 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployer.java
```diff
@@ -175,6 +175,17 @@ public String deploy(AppDeploymentRequest request) {
...
+				// for cross reference to work inside SAJ.
+				// looks like for now we can't remove these env style formats as i.e.
+				// DeployerIntegrationTestProperties in tests really assume 'INSTANCE_INDEX' and
+				// this might be indication that we can't yet fully remove those.
+				if (useSpringApplicationJson(request)) {
+					appInstanceEnv.put("instance.index", Integer.toString(i));
+					appInstanceEnv.put("spring.application.index", Integer.toString(i));
+					appInstanceEnv.put("spring.cloud.application.guid", Integer.toString(port));
+				}
+
```
mminella: Shouldn't the other way of setting properties be in an else here?

jvalkeal: I put a comment to PR to read reason for that from comment in code as I was 100% sure someone would ask it. In nutshell, tests break if we take it away.

mminella: So we should fix the test to be valid under the new conditions.  And since you bring it up, we should have a test for this scenario as well.

jvalkeal: Maybe that `spi-test-app` could be enhanced to track these. We're actually missing integration tests for metric collector as integration with that was only thing what got broken what this change is fixing.

mminella: Why can't we validate these in a similar way to the way we do with the command builders like here: https://github.com/spring-cloud/spring-cloud-deployer-local/blob/master/spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/JavaExecutionCommandBuilderTests.java#L168

jvalkeal: That sounds reasonable. It doesn't test the actual problem for spring resolving cross refs, afaik, but is better than nothing.

markpollack: I agree, should be an `else` cause as it just looks confusing if anyone starts to debug to see the same property set twice (once in env-var and once in SAJ).

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalAppDeployerEnvironmentIntegrationTests.java
```diff
@@ -0,0 +1,258 @@
...
+		return super.testApplication();
+	}
+
+	@Override
+	protected String randomName() {
+		if (LocalDeployerUtils.isWindows()) {
+			// tweak random dir name on win to be shorter
+			String uuid = UUID.randomUUID().toString();
+			long l = ByteBuffer.wrap(uuid.toString().getBytes()).getLong();
+			return name.getMethodName() + Long.toString(l, Character.MAX_RADIX);
```
mminella, line 109: Should we move this up to the parent?  I can do this on merge if you're ok with it.

mminella:
Rebased, squashed, and merged as f417134f334c0c5b21821cac1fa490b480044ff7.

PR #115, opened 2018-04-26
Force INSTANCE_INDEX with SAJ

### LINKED ISSUE #114: Missing proper instance index key
See spring-cloud/spring-cloud-dataflow#2164

jvalkeal:
- For partitioning to work we need to force
  env format as INSTANCE_INDEX into SAJ because
  spring cloud stream uses
  @Value("${INSTANCE_INDEX:${CF_INSTANCE_INDEX:0}}")
- Fixes #114 
- Resolves spring-cloud/spring-cloud-dataflow#2164


markpollack:
what if we set `spring.cloud.stream.instanceIndex` instead?

jvalkeal:
Hmm, I thought I tried that but maybe I messed up something. Theory was that `@Value` might override that as it has a default value `0`. I'll re-try as I might have been wrong.

jvalkeal:
I updated PR to use `spring.cloud.stream.instanceIndex` as it seem to work. I think I originally thought it would not work because I used multiple time sources with rabbit and that would not work. Woodchuck and ticktock with 2 log sinks looked ok.

markpollack:
:+1:

markpollack:
squashed and merged as 602be2151f3eedc092b95cbb76f69752b48132c8

PR #119, opened 2018-05-25
Add debug logging around process shutdown

### LINKED ISSUE #117: Add logging to shutdown procedure
This is just for debugging running this as CI build seem to hang when deployer is trying to shutdown app.

Add debug statetements around rest call and process alive and destroy calls.

jvalkeal:
- Now simply can log things around call for shutdown
  endpoint and interaction for process calls from jdk.
- Fixes #117

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -251,8 +251,10 @@ public abstract class AbstractLocalDeployerSupport {
 		try {
 			int timeout = getLocalDeployerProperties().getShutdownTimeout();
 			if (timeout > 0) {
+				logger.debug("About to call shutdown endpoint for instance {}", instance);
 				ResponseEntity<String> response = restTemplate.postForEntity(
 						instance.getBaseUrl() + "/shutdown", null, String.class);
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -251,8 +251,10 @@ protected void shutdownAndWait(Instance instance) {
 		try {
 			int timeout = getLocalDeployerProperties().getShutdownTimeout();
 			if (timeout > 0) {
+				logger.debug("About to call shutdown endpoint for instance {}", instance);
```
ilayaperumalg, line 254: I think it would be better: for instance -> for the instance. will fix here and in other places when merging.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -270,18 +272,23 @@ protected void shutdownAndWait(Instance instance) {
 		}
 		finally {
 			if (isAlive(instance.getProcess())) {
+				logger.debug("About to call destroy process for instance {}", instance);
 				instance.getProcess().destroy();
+				logger.debug("Call comleted to destroy process for instance {}", instance);
```
ilayaperumalg, line 277: typo: comleted

ilayaperumalg:
Rebased, fixed the comments and merged as eccc853

PR #121, opened 2018-05-25
Configure restTemplate timeouts

### LINKED ISSUE #120: Configure restTemplate to use timeouts
`restTemplate` calling shutdown endpoint is using plain config via its default constructor. On default timeouts are disabled on a client doing http requests.

At minimum, we should use timeouts from `spring.cloud.deployer.local.shutdownTimeout` in http for connection and listen timeouts as that prevents calls to hang up if hang is about to happen.

jvalkeal:
- For template used calling shutdown endpoint, build it
  manually by setting client factory using shutdown timeouts
  for connection and read. This should migitate risks that
  rest call doesn't hang up.
- As default timeout value is 30, this changes default behaviour
  to have a 30 sec timeout. Can be disabled to normal behaviour
  by setting timeout to -1, which then blocks.
- Fixes #120

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -34,6 +34,7 @@ import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
 import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
 import org.springframework.cloud.deployer.spi.util.RuntimeVersionUtils;
 import org.springframework.http.ResponseEntity;
+import org.springframework.http.client.SimpleClientHttpRequestFactory;
 import org.springframework.util.Assert;
 import org.springframework.util.SocketUtils;
 import org.springframework.web.client.RestTemplate;
@@ -67,7 +68,7 @@ public abstract class AbstractLocalDeployerSupport {
 
 	private final LocalDeployerProperties properties;
 
-	private final RestTemplate restTemplate = new RestTemplate();
+	private final RestTemplate restTemplate;
 
 	private final JavaCommandBuilder javaCommandBuilder;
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -85,6 +86,28 @@ public AbstractLocalDeployerSupport(LocalDeployerProperties properties) {
...
+	 * needed can be overridden from an implementing class. This default implementation
+	 * sets connection and read timeouts for {@link SimpleClientHttpRequestFactory} and
+	 * configures {@link RestTemplate} to use that factory. If shutdown timeout in
+	 * properties negative, returns default {@link RestTemplate} which doesn't use timeouts.
+	 *
+	 * @param properties the local deployer properties
+	 * @return the rest template
+	 */
+	protected RestTemplate buildRestTemplate(LocalDeployerProperties properties) {
+		if (properties != null && properties.getShutdownTimeout() > -1) {
```
markpollack, line 103: Is zero a valid timeout value?  Maybe we can add `@Min(0)` on the property itself?

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalDeployerSupportTests.java
```diff
@@ -93,6 +93,32 @@ public void testCalcServerPort() throws MalformedURLException {
 		assertThat(portToUse, not(9292));
 	}
 
+	@Test
+	public void testShutdownPropertyConfiguresRequestFactory() throws Exception {
+		LocalDeployerProperties properties = new LocalDeployerProperties();
+		properties.setShutdownTimeout(1);
+		AbstractLocalDeployerSupport abstractLocalDeployerSupport = new AbstractLocalDeployerSupport(properties) {};
+		Object restTemplate = TestUtils.readField("restTemplate", abstractLocalDeployerSupport);
```
markpollack, line 101: Think we can use `ReflectionTestutils` in the spring-test package instead of our own util class.

```
		Object restTemplate = ReflectionTestUtils.getField(abstractLocalDeployerSupport, "restTemplate");
		Object requestFactory = ReflectionTestUtils.getField(restTemplate, "requestFactory");
		Object connectTimeout = ReflectionTestUtils.getField(requestFactory, "connectTimeout");
		Object readTimeout = ReflectionTestUtils.getField(requestFactory, "readTimeout");```

will check other usages.

markpollack:
merged as 4a89ba72185db9d29b050b5b93102e7d2c9d1b6d

PR #123, opened 2018-06-06
Support zero port

### LINKED ISSUE #122: Support using zero port
Currently local deployer totally fails as it tries to connect to defined port. We should not just return deploying if zero port is used, instead just return deployed if process is running. This at least allows user to use a real random port.

jvalkeal:
- Now simply handling case where we don't have
  a known port by returning `deployed` if process
  is alive. With fully random port, deployer fully
  fail to understand anything about trying to open
  http connection.
- This is best we can do for now as it's usually
  convenient to use zero ports.
- Fixes #122


ilayaperumalg:
LGTM, merging.

PR #126, opened 2018-07-11
Prune task instance history map

### LINKED ISSUE #112: Manage active Task Map
A cleanup routine would be in place to make sure the Task PIDs in the `running` Map doesn't grow indefinitely. 

Following two items are out of scope for this issue.

~We would have a separate thread to double check whether the PID is still alive; if not, we will remove it from the Map.~

~Also, when a cleanup/destroy routine is invoked from SCDF, we will make sure to cleanup the PID from running Map. This is already in place for Streams; we would implement it for Tasks now.~

tzolov:
This solution aim to clean the history of repeatedly executed tasks. Only the last launched Task instance is kept in the running map.
- Track the previous task-launch instance for the any given Task (identified by the Task Name)
- On new Task launch event, find all previous task instances for the Task being launched.
- Found instances in non-running and non-launching state are removed from the running and the tracking maps.

Resolves #112

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
@@ -29,6 +29,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
 import javax.annotation.PreDestroy;
 
 import org.slf4j.Logger;
@@ -50,6 +52,7 @@ import org.springframework.cloud.deployer.spi.task.TaskStatus;
  * @author Thomas Risberg
  * @author Oleg Zhurakousky
  * @author Michael Minella
+ * @author Christian Tzolov
  */
 public class LocalTaskLauncher extends AbstractLocalDeployerSupport implements TaskLauncher {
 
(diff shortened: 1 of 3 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
```diff
@@ -111,6 +118,26 @@ public String launch(AppDeploymentRequest request) {
 		return taskLaunchId;
 	}
 
+	private void pruneTaskInstanceHistory(String taskDefinitionName, String taskLaunchId) {
```
cppwfs, line 121: The only issue I see is that as we would be returning false `unknown` status for tasks that have been purged.   i.e. if I hit the /runtime/apps/id I will get `unknown`s for tasks that have successfully completed(then purged).  And we will have at most only one `complete` entry at a time for tasks.  Maybe this is a @sabbyanandan question.   Do people use this to check the status of running tasks?   Or should they be referred to the /tasks/executions endpoint?

sabbyanandan: We have had reports of status being inconsistent in the list, summary, and executions pages. So, yeah, we have users using the different options. However, the "runtime" tab would show what is running, though. Do we show the tasks in that tab even if they are destroyed?

cppwfs: Just took a look at the controller, only the streamDeployer is used to return statuses from the runtime/apps.   So I guess the question is, "Does anyone really use this feature in the local deployer?"

cppwfs:
Had a chat with @mminella on the issue I raised.    LGTM

cppwfs:
Squashed, Rebased Merged.

PR #130, opened 2018-11-12
Improve random port selection

### LINKED ISSUE #129: Improve random port selection
Acceptance tests are failing often due to binding conflicts.

markpollack:
Fixes #129

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -18,7 +18,10 @@ package org.springframework.cloud.deployer.spi.local;
 
 import java.io.IOException;
 import java.net.URL;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Random;
@@ -55,6 +58,12 @@ import org.springframework.web.client.RestTemplate;
  * @author Michael Minella
  */
 public abstract class AbstractLocalDeployerSupport {
+
+	protected static Set<Integer> usedPorts = Collections.newSetFromMap(new LinkedHashMap<Integer, Boolean>(){
+		protected boolean removeEldestEntry(Map.Entry<Integer, Boolean> eldest) {
+			return size() > 1000;
+		}
+	});
 
 	protected final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -368,7 +377,7 @@ public abstract class AbstractLocalDeployerSupport {
 		Integer commandLineArgPort = isServerPortKeyPresentOnArgs(request);
 
 		if(useDynamicPort) {
-			port = getRandomPort(port);
+			port = getRandomPort();
 		}
 		else if(commandLineArgPort != null) {
 			port = commandLineArgPort;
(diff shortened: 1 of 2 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -384,23 +393,42 @@ else if(request.getDefinition().getProperties().containsKey(LocalAppDeployer.SER
...
+	public synchronized int getRandomPort() {
+		Set<Integer> availPorts = new HashSet<>();
+		// SocketUtils.findAvailableTcpPorts retries 6 times, add additional retry on top.
+		for (int retryCount = 0; retryCount < 5; retryCount++) {
+			int randomInt = ThreadLocalRandom.current().nextInt(20000, 61000);
+			try {
+				availPorts = SocketUtils.findAvailableTcpPorts(5, randomInt, randomInt + 5);
+				try {
+					// Give some time for the system to release up ports that were scanned.
+					Thread.sleep(100);
```
jvalkeal, line 405: I don't think we need to sleep here. If this is because I commented something it was just a theoretical question.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -56,6 +59,12 @@
  */
 public abstract class AbstractLocalDeployerSupport {
 
+	protected static Set<Integer> usedPorts = Collections.newSetFromMap(new LinkedHashMap<Integer, Boolean>(){
+		protected boolean removeEldestEntry(Map.Entry<Integer, Boolean> eldest) {
```
jvalkeal, line 63: Would like to see `@Override` here

jvalkeal:
Other than that lgtm.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
markpollack: I had a more instrumented test harness counting all the ways stuff could fail, I saw the failure rate go down when the sleep was added.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
markpollack: :taco: and :+1:   (was looking for `:thumbsup` and found taco

mminella:
Should we consider providing feedback or an issue around `SocketUtils` to the framework team?  As a user of `SocketUtils`, based on the javadoc, I wouldn't expect to need to do the extra work we're having to do here.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
jvalkeal: Well that sucks. It means that this concept is even more unreliable what I thought.

jvalkeal:
@mminella it was a mistake that `SocketUtils` even got copied from SI's test classes to framework.

PR #131, opened 2018-11-15
Nano instead of milliseconds in the name of the local log folder

eutkin:
Hi all. We run task faster then 1 millisecondsi

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
@@ -16,6 +16,15 @@
 
 package org.springframework.cloud.deployer.spi.local;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.cloud.deployer.spi.core.AppDeploymentRequest;
+import org.springframework.cloud.deployer.spi.core.RuntimeEnvironmentInfo;
+import org.springframework.cloud.deployer.spi.task.LaunchState;
+import org.springframework.cloud.deployer.spi.task.TaskLauncher;
+import org.springframework.cloud.deployer.spi.task.TaskStatus;
+
+import javax.annotation.PreDestroy;
 import java.io.File;
 import java.io.IOException;
 import java.net.HttpURLConnection;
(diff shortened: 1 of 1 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
```diff
@@ -205,7 +204,8 @@ private Path createLogDir(AppDeploymentRequest request) throws IOException {
 							request.getDefinition().getName());
 		}
 
-		String qualifiedName = Long.toString(System.currentTimeMillis());
+		Instant now = Instant.now();
```
jvalkeal: Why u didn't use `System.nanoTime()`?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
```diff
@@ -205,7 +204,8 @@ private Path createLogDir(AppDeploymentRequest request) throws IOException {
...
+		String qualifiedName = Long.toString(now.toEpochMilli() + now.getNano());
```
jvalkeal: Btw, this will not actually work as it has similar precision issue:
```
    	long t1 = System.currentTimeMillis();
    	long t2 = System.currentTimeMillis();
    	long t3 = System.nanoTime();
    	long t4 = System.nanoTime();

    	System.out.println(Long.toString(t1));
    	System.out.println(Long.toString(t2));
    	System.out.println(Long.toString(t3));
    	System.out.println(Long.toString(t4));

    	Instant now1 = Instant.now();
    	Instant now2 = Instant.now();
    	long t5 = now1.toEpochMilli() + now1.getNano();
    	long t6 = now2.toEpochMilli() + now2.getNano();
    	System.out.println(Long.toString(t5));
    	System.out.println(Long.toString(t6));
```
which results:
```
1542298923798
1542298923798
1752380359022364
1752380359022472
1543097923799
1543097923799

```

jvalkeal:
Thx buddy, I'm gonna merge this.

jvalkeal:
Merged per f6943ebabcc3852cfb3ebee0f65a2c5d7c04c34d

PR #134, opened 2019-03-15
Configurable random port range

### LINKED ISSUE #133: Make the random-port range configurable
Currently if the `server.port` is not set via a property, the application instance's port is computed randomly from the fixed range `(20000, 61000)`. 

Certain use cases such as the prometheus monitoring, necessitate access to the instance's actuator entry points. When deployed in a docker container with current, hardcoded port range (20000, 61000) it is practically impossible to expose and share those apps with external services (such as prometheus) running in another container. 

The ability to set the port range spam would allow us to to have a smaller (and exportable) range that satisfies above use-case.

tzolov:
- Add spring.cloud.deployer.local.portBound.lower and spring.cloud.deployer.local.portBound.upper property to configure the range of the generated random properties.
  - lower defaults to 20000 and upper to 61000 (e.g. the previous fixed values).
  - The random port is generated in the range [lower, upper + 5]

  Resolves #133

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
@@ -122,6 +122,45 @@ public class LocalDeployerProperties {
 	 */
 	private boolean useSpringApplicationJson = true;
 
+	private PortBound portBound = new PortBound();
+
+	public static class PortBound {
+
+		/**
+		 * Lower bound for computing applications's random port.
+		 */
+		private int lower = 20000;
+
+		/**
+		 * Lower bound for computing applications's random port.
+		 */
+		private int upper = 61000;
+
+		public int getLower() {
+			return lower;
+		}
+
+		public void setLower(int lower) {
+			this.lower = lower;
+		}
+
+		public int getUpper() {
+			return upper;
+		}
+
+		public void setUpper(int upper) {
+			this.upper = upper;
+		}
+
+		@Override
+		public String toString() {
+			return "{" +
+					"lower=" + lower +
+					", upper=" + upper +
+					'}';
+		}
+	}
+
 	public String getJavaCmd() {
 		return javaCmd;
 	}
@@ -179,6 +218,10 @@ public class LocalDeployerProperties {
 		this.useSpringApplicationJson = useSpringApplicationJson;
 	}
 
+	public PortBound getPortBound() {
+		return portBound;
+	}
+
 	private String deduceJavaCommand() {
 		String javaExecutablePath = JAVA_COMMAND;
 		String javaHome = System.getProperty("java.home");
@@ -198,7 +241,7 @@ public class LocalDeployerProperties {
 	}
 
 	@Override
-	public String toString(){
+	public String toString() {
 		return new ToStringCreator(this)
 				.append("workingDirectoriesRoot", this.workingDirectoriesRoot)
 				.append("javaOpts", this.javaOpts)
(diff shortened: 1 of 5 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -122,6 +122,45 @@
 	 */
 	private boolean useSpringApplicationJson = true;
 
+	private PortBound portBound = new PortBound();
```
jvalkeal: These feels a bit awkward as is:
```
spring.cloud.deployer.local.port-bound.upper
spring.cloud.deployer.local.port-bound.lower
```
would it be better to rename to:
```
spring.cloud.deployer.local.port-range.high
spring.cloud.deployer.local.port-range.low
```

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/BoundedRandomPortTests.java
```diff
@@ -0,0 +1,49 @@
...
+	@Before
+	public void setUp() {
+		LocalDeployerProperties properties = new LocalDeployerProperties();
+		properties.getPortBound().setLower(30001);
+		properties.getPortBound().setUpper(30213);
+		localDeployerSupport = new AbstractLocalDeployerSupport(properties) {};
+	}
+
+	@Test
+	public void portTests() {
```
jvalkeal: It'd be good to do actual test to bind using a real property key/value as that verifies that boot binds things ok.

tzolov: Makes sense, i will add integration test for this. 

Not sure about the Docker mode tough? are we testing this?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
tzolov: agree

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/BoundedRandomPortTests.java
jvalkeal: this was just about testing boot binding and it really have nothing to do with docker.

jvalkeal: Random ports really doesn't work with docker if you need to expose something. But then we're not using random ports with docker?

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/RandomPortRangeTests.java
```diff
@@ -0,0 +1,48 @@
...
+		LocalDeployerProperties properties = new LocalDeployerProperties();
+		properties.getPortRange().setLow(30001);
+		properties.getPortRange().setHigh(30213);
+		localDeployerSupport = new AbstractLocalDeployerSupport(properties) {};
+	}
+
+	@Test
+	public void portTests() {
+		for (int i = 0; i < 30; i++) {
+			System.out.println(i);
```
jvalkeal: println could go away

tzolov: addressed with the commit below

jvalkeal:
Merged per dc2ad300800cb5cf39da6981699351ebeb7136f2

PR #147, opened 2019-03-22
Implement new methods in LocalTaskLauncher

dturanski:
Part of https://github.com/spring-cloud/spring-cloud-dataflow/issues/2855 

Depends on https://github.com/spring-cloud/spring-cloud-deployer/pull/286


spring-cloud-deployer-local/pom.xml
```diff
@@ -43,6 +43,11 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-validation</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.hibernate.validator</groupId>
+			<artifactId>hibernate-validator-annotation-processor</artifactId>
+			<version>6.0.2.Final</version>
```
jvalkeal: Version should go to parent's dependency management. Don't we get any dependency management for this from boot/cloud builds?

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalDeployerPropertiesTests.java
```diff
@@ -0,0 +1,52 @@
...
+properties = {"spring.cloud.deployer.local.maximumConcurrentTasks=2",
+	"spring.cloud.deployer.local.platformMaximumConcurrentTasks=1"})
+
+@RunWith(SpringRunner.class)
+public class LocalDeployerPropertiesTests {
+	@Autowired LocalDeployerProperties deployerProperties;
+	@Test
+	public void testMaxConcurrentPropeties() {
+
+		//deployerProperties.setShutdownTimeout(-999);
```
jvalkeal: Did you forget to actually implement boot binding tests?

jvalkeal: Check i.e. `RandomPortRangeContextTests` how boot makes it easy to test these bindings.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
```diff
@@ -177,6 +185,23 @@ public RuntimeEnvironmentInfo environmentInfo() {
...
+	@Override
+	public int getMaximumConcurrentTasks() {
+		return getLocalDeployerProperties().getMaximumConcurrentTasks();
+	}
+
+
+	@Override
+	public int getRunningTaskExecutionCount() {
+		return
+			new Long(running.entrySet().stream().filter( e-> e.getValue().getProcess().isAlive()).count()).intValue();
```
jvalkeal: It definitely feels a bit weird to return Long when signature has int.

spring-cloud-deployer-local/src/test/java/org/springframework/cloud/deployer/spi/local/LocalDeployerPropertiesTests.java
dturanski: Whoops. This test was left over from a previous incarnation. It's no longer needed

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
dturanski: The compiler would agree, must return an int.  Using `long Stream.count()`, so I'm converting it to int using `Long.intValue()`

jvalkeal: Ah yes, I missed to follow all the brackets.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
```diff
@@ -78,6 +79,13 @@ public LocalTaskLauncher(LocalDeployerProperties properties) {
 	@Override
 	public String launch(AppDeploymentRequest request) {
 
+		if (this.maxConcurrentExecutionsReached()) {
+			throw new IllegalStateException(
+				String.format("Cannot launch task %s. The maximum concurrent task executions is at its limit [%d].",
+					request.getDefinition().getName(), this.getLocalDeployerProperties().getMaximumConcurrentTasks())
+			);
+		}
```
mminella, line 87: While I'm not advocating for anything other than this now, we will need to consider how to make this more robust in the future (queuing mechanism, etc).

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalTaskLauncher.java
mminella: It's my understanding that it is considered bad practice to use streams where a simple `for` loop can be used given the performance implications (basically, streams are slower and more heavy weight).  Ollie mentioned it in this thread as an example: https://twitter.com/odrotbohm/status/1056866589763690497

dturanski:
I'll change it. I don't like the way it looks anyway.

markpollack:
squashed and merged as bb58c04b70496c4bd2284138b502a6f93c227c7f

PR #161, opened 2019-06-27
Fix debug and inherit settings

### LINKED ISSUE #160: Debug flags not exposed as deployment properties
Currently you can set things like
```
deployer.log.local.debugPort=8889
deployer.log.local.debugSuspend=n
deployer.time.local.debugPort=8888
deployer.time.local.debugSuspend=n
```
However these are just manually processed and UI dialog doesn't show these as they are missing from generated metadata. Also you can't use other formats like `debug-port`.

Cases for `debugPort` vs. `debug-port` is a other related issue as in a past we started to use deployment properties manually without doing a proper binding to `LocalDeployerProperties` we lost that automatic handling or `kebab-case` vs. `camelCase`. This may very well be issue all over the places, not just local deployer.

jvalkeal:
- Change original keys debugPort, debugSuspend and inheritLogging to
  a proper kebab-case keys, debug-port, debug-suspend and inherit-logging.
- Fix LocalDeployerProperties so that these keys can be properly binded
  and that keys actually go into boot metadata.
- When these are used from a deployment request, properly use boot's
  property Binder so that both camelCase and kebab-case work.
- Add debug-suspend to a additional-spring-configuration-metadata.json with
  hints `y` and `n` so that those will be picked as option set in UI.
- Various tests to see that all works.
- This commit will leave other properties use of camelCase vs. kebab-case
  unchanged as it's a bit larger topic.
- Fixes #160

NOTE: This will then break some skipper tests as there's 3 new deployment properties and we seem to check count returned for local deployer.

Here's example how these new properties will show up in UI:

local-deployer-debug-options

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
@@ -33,7 +33,9 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
+import org.springframework.boot.context.properties.bind.Bindable;
+import org.springframework.boot.context.properties.bind.Binder;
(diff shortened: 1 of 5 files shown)


jvalkeal:
@ilayaperumalg when you get to this, there's a PR in skipper to fix tests what this PR will cause.

ilayaperumalg:
LGTM, merging.

PR #175, opened 2019-11-19
Bind Java execution properties

### LINKED ISSUE #174: Some properties not bind properly
For example we need to use `deployer.timestamp.local.javaOpts` as `deployer.timestamp.local.java-opts` doesn't work because we don't do proper boot binding.

ilayaperumalg:
- The AppDeployment request's deployment properties need to be bound correctly based on the Spring Boot Binder to allow multiple possible key options
 - Fix javaCmd, javaOpts local deployer properties to be handled correctly
 - Remove unused main, classpath handling
 - Add tests

Resolves #174


jvalkeal:
Works great, merging...

PR #186, opened 2020-05-22
Run apps as docker containers in docker-compose

tzolov:
- The spring.cloud.deployer.local.dockerNetwork allows setting the container network. When provided it adds --network <scdf network name> to the DockerCommandBuilder to ensure that the deployed app containers are connected to the network created by the docker-compose (note that the -p project-name). For default docker-compose installations set the dockerNetwork to spring-cloud-dataflow-server_default
 - Set the deployment-id as docker container name of the deployed app.
 - Change the AppInstance URL to <deployement-id>:<app-port> (the URL uses the local/skipper IP otherwise).
 - Backward compatible with not docker compose local installations

Related to: https://github.com/spring-cloud/spring-cloud-dataflow/issues/3966
Related to: https://github.com/spring-cloud/spring-cloud-dataflow/issues/3643

Based on: https://docs.google.com/document/d/1gZrqF6QCVqdYVJ8LAModPr_O0dJUSGW-KSFAI1tuYcE/edit?usp=sharing

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
@@ -170,6 +170,16 @@ public class LocalDeployerProperties {
 
 	private boolean inheritLogging;
 
+	private String dockerNetwork = "bridge";
+
+	public String getDockerNetwork() {
+		return dockerNetwork;
+	}
+
+	public void setDockerNetwork(String dockerNetwork) {
+		this.dockerNetwork = dockerNetwork;
+	}
+
 	public Integer getDebugPort() {
 		return debugPort;
 	}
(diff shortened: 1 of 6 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -170,6 +170,16 @@ public String toString() {
 
 	private boolean inheritLogging;
 
+	private String dockerNetwork = "bridge";
+
+	public String getDockerNetwork() {
+		return dockerNetwork;
+	}
+
+	public void setDockerNetwork(String dockerNetwork) {
```
jvalkeal: Would it be better if we'd start to group these settings, like `spring.cloud.deployer.local.docker.network`. Might look more clean especially if we add more settings.

tzolov: makes sense. Just changed it into docker.network and added some related tests.

jvalkeal:
Merged per 71a8153b2c46ba10b35ebb5af274724eab6d7b08

PR #192, opened 2020-08-14
Add startup/health probes for stream apps

### LINKED ISSUE #189: Support better probes.
It would be good to have support for app `startup` and `health` http probes. Main motivation is to make an app _fail_ or never come _deployed_ which makes testing much easier and then you can demonstrate some stream update failures locally as well.

jvalkeal:
- Adding an optional startup probe which simply checks
  2xx return code to mark when new app goes from deploying
  to deployed.
- Adding an optional health probe which simply checks
  2xx return code to mark when running app is marked either
  deployed or failed.
- Essentially adding `startup-probe.path` and `health-probe.path`
  with very minimal functionality but allows to easily add
  more features in future if needed.
- Together these allow to do failed stream update locally as
  previously it has only been possible in cf/k8s.
- Fixes #189

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/HttpProbeExecutor.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/HttpProbeExecutor.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/HttpProbeExecutor.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.DefaultUriBuilderFactory;
+
+public class HttpProbeExecutor {
+
+	private static final Logger logger = LoggerFactory.getLogger(HttpProbeExecutor.class);
+    private final RestTemplate restTemplate;
+    private final URI uri;
+
+    public HttpProbeExecutor(RestTemplate restTemplate, URI uri) {
```
ilayaperumalg, line 42: Please add author tag... so we know who to blame ;)

ilayaperumalg:
LGTM, merging.

PR #195, opened 2020-08-26
Use both stderr and stdout

### LINKED ISSUE #193: Getting log defaults to stderr
We get content from `stderr` before `stdout` but recently some tests have started to fail on a dataflow side, This is cause by some environments logging `Picked up JAVA_TOOL_OPTIONS` in stderr when `JAVA_TOOL_OPTIONS` is set:
```
2020-08-21T15:30:47.2695234Z [ERROR] Failures: 
2020-08-21T15:30:47.2696373Z [ERROR]   TaskCommandTests.testGetLog:138 Task execution log = CommandResult [success=true, result=Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
```
This kinda shows flaw in an idea that stderr always shows a real error.

We need to change this logic to be more better.

jvalkeal:
- When getting logs for stream and task apps we used
  to just read stderr if it had some content which
  is a bit problematic as stderr may have some content
  which is not a real error.
- This i.e. happens with JAVA_TOOL_OPTIONS env variable.
- Now simply getting both.
- Fixes #193


jvalkeal:
Looks like this fixes tests, merging.

PR #198, opened 2020-09-23
Support remote debug host:port syntax for Java 9 and above

dturanski:


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -474,6 +482,30 @@ protected Integer isServerPortKeyPresentOnArgs(AppDeploymentRequest request) {
 		return result;
 	}
 
+	static class JavaVersionParser {
+		private final String versionAsString;
+
+		JavaVersionParser() {
+			this(System.getProperty("java.specification.version"));
```
tzolov: Is the `java.specification.version` present on non-Oracle JDKs?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java
```diff
@@ -474,6 +482,30 @@ protected Integer isServerPortKeyPresentOnArgs(AppDeploymentRequest request) {
...
+		}
+
+		JavaVersionParser(String versionAsString) {
+			this.versionAsString = versionAsString;
+		}
+
+		//Earlier java versions 1.8 and below use '1.x'. Later vesions use 'X'
+		int versionAsOrdinal() {
+			if (versionAsString.startsWith("1.")) {
```
tzolov: Perhaps we should handle null/empty `versionAsString` value (e.g. when the `System.getProperty("java.specification.version")` doesn't have value for whatever reasons.

tzolov (commented):
I started testing with the old (e.g. Einstein) `time source` app, using the latest skipper:2.6.0-SH and scdf:2.7.0-SH that run inside the new jdk11 baseimages.   The `time` app crashes on start with error:
```
Picked up JAVA_TOOL_OPTIONS: -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=*:20075
ERROR: transport error 202: gethostbyname: unknown host
ERROR: JDWP Transport dt_socket failed to initialize, TRANSPORT_INIT(510)
JDWP exit error AGENT_ERROR_TRANSPORT_INIT(197): No transports initialized [debugInit.c:750]
```
from the `*:20075` address i assume that the local deployer detects the jdk11 runtime, but this is the runtime of the Skipper, while my applications runs in its own docker container (which is still jdk8 based) . 

Current configurations approach doesn't allow me to override the the jdk11 selection.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DebugAddress.java
```diff
@@ -0,0 +1,114 @@
...
+
+	public String getSuspend() {
+		return suspend;
+	}
+
+	public String getAddress() {
+		return this.address;
+	}
+
+	public String getDebugCommand() {
```
dturanski: I would remove this method from this class. Single responsibility principle.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
```diff
@@ -61,24 +64,47 @@ public DockerCommandBuilder(String dockerNetwork, boolean deleteContainerOnExit)
...
 		commands.add("docker");
 		commands.add("run");
+
 		if (StringUtils.hasText(this.dockerNetwork)) {
 			commands.add("--network");
 			commands.add(this.dockerNetwork);
 		}
-		if (this.deleteContainerOnExit) {
+
+		if (this.deleteContainerOnExit && localDeployerProperties.getDocker().isDeleteContainerOnExit()) {
```
dturanski: I'm not following this. Should it be `||` ?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
```diff
@@ -289,18 +329,26 @@ public Integer getDebugPort() {
...
 	}
 
-	public void setDebugSuspend(String debugSuspend) {
+	public void setDebugSuspend(DebugSuspendType debugSuspend) {
 		this.debugSuspend = debugSuspend;
 	}
 
 	public void setDebugPort(Integer debugPort) {
 		this.debugPort = debugPort;
 	}
```
dturanski, line 359: Can we log a warning here `'port' property is deprecated. Use 'address' instead`.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DebugAddress.java
tzolov: I will move it as a default method: `getJdwpOptions(suspend, address)` inside the CommandBuilder interface.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/LocalDeployerProperties.java
tzolov: done

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DockerCommandBuilder.java
tzolov: The logic is that you can disable container-deletion-on-exit for all containers (e.g. `this.deleteContainerOnExit`) or for a singe app container (`localDeployerProperties.getDocker().isDeleteContainerOnExit()`). 

In current implementation the first is configured at [local deployer boot time](https://github.com/spring-cloud/spring-cloud-deployer-local/blob/master/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/AbstractLocalDeployerSupport.java#L106-L107) e.g. you app you this with DataFlow's `SPRING_CLOUD_DATAFLOW_TASK_PLATFORM_LOCAL_ACCOUNTS_DEFAULT_DOCKER_DELETE_CONTAINER_ON_EXIT`  or Skipper's `SPRING_CLOUD_SKIPPER_SERVER_PLATFORM_LOCAL_ACCOUNTS_DEFAULT_DOCKER_DELETE_CONTAINER_ON_EXIT` properties. 
Then you can use the pipeline deployment property to set the desired behavior for a particular app at the time of deployment. 

In other words you can set the deletion to true by default, but have the choice to alter this when you deploy the pipeline, which is handy to check the logs of some failing app without having to  restart/reconfigure the SCDF&Skipper.

dturanski: So if I ask for a specific container to be deleted, I expect that would override the default setting of false.

tzolov: indeed

dturanski: this.deleteOnExit=false
this.deployerProperties.getDocker().isDeleteContainerOnExit()=true
then `true && false == false`. So no `--rm`.  

So you can delete all but one pipeline, or you can delete every pipeline, or you don't delete anything,  but you can't delete only one pipeline.  I'm not sure if that makes sense.  I don't think it's possible to handle every case unless `deleteContainerOnExit` is default null, or `Optional.empty()` .

tzolov: Note that `this.deployerProperties.getDocker().isDeleteContainerOnExit()=true` is per application **not** per stream/pipeline! So in this case the `no --rm` means that all other apps in the pipeline will be deleted but this one will be kept so you check the logs or inspect its environment. 
I've already used it for troubleshooting a failing app. IMO It is easier if you try it. Use `watch docker ps -a` to monitor the behavior of the apps in the pipeline after you deploy or un-deploy it.

tzolov: You can think of `this.deleteOnExit` as policy for the entire pipeline, while the `this.deployerProperties.getDocker().isDeleteContainerOnExit()` is per single app in that pipeline.
By default the this.deleteOnExit is true because in general case you don't want to keep all containers for all pipelines. Thinking about this perhaps `this.deleteOnExit=false` e.g. keep all containers for all pipelines is never desired? Perhaps using the this.deployerProperties.getDocker().isDeleteContainerOnExit() for cherry pinking what containers to stay would be enough?

tzolov: I will remove the `this.deleteOnExit`

tzolov: https://github.com/spring-cloud/spring-cloud-deployer-local/pull/198/files#diff-9984d89909b44a5e9c80671c6cafb573R105-R107

PR #210, opened 2021-10-11
Prevent logging Env. Variables

tzolov:
The way that Task (And Task Launcher) handle and use env. variables brings hight risk to expose
   sensitive information. Therefore it is safer not to list variables's values in any log level.


jvalkeal:
Looks like that was merged per b54f5e5afc1e1cb8f152da032a14a41f5e54a186, and we forget to close.

PR #214, opened 2022-03-09
To support Spring Boot 3.0 the local deployer needs util to find open port.

### LINKED ISSUE #213: SocketUtils has been removed from Spring Framework 6.x
It has been deprecated since 5.3.  So now when a user tries to use a LocalDeployer on Boot 3.x it fails stating that SocketUtils is missing.  Local Deployer needs to include a DeployerSocketUtils library so that it can find an open port.

cppwfs:
This is because SocketUtils was removed from the Spring Framework.

resolves #213

### DIFF
--- a/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
+++ b/spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
@@ -1,0 +1,296 @@
+/*
+ * Copyright 2016-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.deployer.spi.local;
+
+import java.net.DatagramSocket;
+import java.net.InetAddress;
+import java.net.ServerSocket;
+import java.util.Random;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import javax.net.ServerSocketFactory;
+
+import org.springframework.util.Assert;
+
+/**
+ * Simple utility methods for working with network sockets &mdash; for example, for
+ * finding available ports on {@code localhost}.
+ *
+ * <p>
+ * Within this class, a TCP port refers to a port for a {@link ServerSocket}; whereas, a
+ * UDP port refers to a port for a {@link DatagramSocket}.
+ *
+ * <p>
+ *
+ * @author Sam Brannen
+ * @author Ben Hale
+ * @author Arjen Poutsma
+ * @author Gunnar Hillert
+ * @author Gary Russell
+ * @since 4.0
+ */
+public abstract class DeployerSocketUtils {
+
+	/**
+	 * The default minimum value for port ranges used when finding an available socket
+	 * port.
+	 */
+	public static final int PORT_RANGE_MIN = 1024;
+
+	/**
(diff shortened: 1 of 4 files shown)


spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
```diff
@@ -0,0 +1,296 @@
...
+import javax.net.ServerSocketFactory;
+
+import org.springframework.util.Assert;
+
+/**
+ * Simple utility methods for working with network sockets &mdash; for example, for
+ * finding available ports on {@code localhost}.
+ *
+ * <p>
+ * Within this class, a TCP port refers to a port for a {@link ServerSocket}; whereas, a
```
onobc: This looks like a lift/shift from the now removed `SocketUtils`. If so, we should probably state that somewhere in the comments. Otherwise, the copyrights/authors list raises an eyebrow. 

Also, the original reason for it being [removed from SF](https://github.com/spring-projects/spring-framework/issues/28052)

> However, these utilities make no guarantee about the subsequent availability of a given port and are therefore unreliable.

although it has been reliable for our uses so far, correct?

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
```diff
@@ -0,0 +1,296 @@
...
+	 * Find the requested number of available TCP ports, each randomly selected from the
+	 * range [{@code minPort}, {@code maxPort}].
+	 * @param numRequested the number of available ports to find
+	 * @param minPort the minimum port number
+	 * @param maxPort the maximum port number
+	 * @return a sorted set of available TCP port numbers
+	 * @throws IllegalStateException if the requested number of available ports could not
+	 * be found
+	 */
+	public static SortedSet<Integer> findAvailableTcpPorts(int numRequested, int minPort, int maxPort) {
```
onobc, line 92: I am in favor of ditching all of the unused API footprint. Lets only keep what is absolutely necessary. Wdyt?

onobc (changes requested):
Hi @cppwfs - thanks for getting ahead of this one. It looks good. I do have a couple requests inline though.

garyrussell:
FYI, this technique is unreliable - hence the removal from SF; it is possible that the port may be in use by the time you start listening on it.

The user of this needs to defend against that and try again.

cppwfs:
@garyrussell Agreed. This PR, was to restore functionality that was lost because of the removal of SocketUtils and restore the functionality so that affected downstream projects' functionality can be restored.  I will create an issue that suggests we use ServerSocket to identify available ports.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
cppwfs: Agreed.  I removed all unessential methods and created https://github.com/spring-cloud/spring-cloud-deployer-local/issues/215 to enhance this behavior.

spring-cloud-deployer-local/src/main/java/org/springframework/cloud/deployer/spi/local/DeployerSocketUtils.java
```diff
@@ -31,266 +30,148 @@
...
- * UDP port refers to a port for a {@link DatagramSocket}.
- *
- * <p>
+ * This is a replacement for SocketUtils.   SocketUtils was introduced in Spring Framework 4.0,
+ * primarily to assist in writing integration tests which start an external server on an available random port.
+ * However, these utilities make no guarantee about the subsequent availability
+ * of a given port and are therefore unreliable. Instead of using SocketUtils to
+ * find an available local port for a server, it is recommended that you rely on a
+ * server's ability to start on a random port that it selects or is assigned by the operating system.
+ * To interact with that server, you should query the server for the port it is currently using.
```
onobc, line 39: If we immediately deprecate this it will also nag us into the reminder that we need to replace it. Wdyt?

onobc (approved):
Thanks for the updates and creating the subsequent follow-on ticket.

onobc:
Closed via https://github.com/spring-cloud/spring-cloud-deployer-local/commit/6cfd6c74a7eed07883d378f6cc9c0cd4fa9974ad

